PR #23, opened 2019-01-20
Fix spelling

nurikabe:
Spelling error for the Docker config.

Speaking of which, could you elaborate on how this works? By "point to location where docker script is found", does that mean the Docker executable (eg. `/usr/bin/docker`), or `mainnet.sh`, part of the Tezos Docker distribution?


habanoz:
Merged. I also update the help output. This is the location where tezos docker script is found. For example for mainnet it should be the location at which mainnet.sh is kept.

PR #28, opened 2019-02-01
Minimum delegation requirement

utdrmac:
Support not paying rewards to those that delegate less than the bakers require minimum. Right now, any reward that would have been sent to skipped delegator is returned to baker and not added to the total reward pool.


habanoz:
Merging.

Issue #29, opened 2019-02-01
Transfer fee is taken from delegators


utdrmac:
**Describe the bug**
Delegator is awarded Y XTZ. When the transfer operation is created, the fee is deducted from Y. Delegator receives (Y - txnFee).

**Expected**
It is expected that delegators receive Y as their calculated reward.

**Implementation**
Give bakers the option to pay the transfer fee if they wish.

habanoz:
Agreed. Merging and closing.

PR #30, opened 2019-02-01
Allow bakers to decide who pays transfer fee

utdrmac:
Default in fee.ini is 'true' to preserve pre-existing behavior


utdrmac:
Resolves #29

Issue #64, opened 2019-03-19
Total Rewards Calculation and Denunciation


amzid:
In the following calculation of the total rewards (to be found in tzscan/tzscan_reward_calculator.py): ![total_rewards_calculation](https://user-images.githubusercontent.com/33556352/54607145-ecf61c00-4a4d-11e9-8f99-6c230a3fbae3.png)
I expected the **lost rewards from denunciation** and **lost fees from denunciation** to be subtracted from the total amount of rewards.

And also if e.g. due to a double baking process, the baker looses the frozen deposits in that cycle, so that ultimately the total gain from that cycle (for the balance of the baker) is negative, should the lost deposits be considered in the calculation of the total rewards (i.e. subtract one more term which is **lost_deposit_from_denunciation**) or should the total amount of rewards be completely independent of any potential lost deposits?

habanoz:
My view is, lost deposits should not be part of total rewards while lost rewards/fees should be. 

Because, I believe loss of deposit is completely at the responsibility of the baker. Delegation is the act of leaving the risk to the baker at the cost of some portion of the baking reward.

habanoz:
Code merged. Closing.

PR #65, opened 2019-03-21
Get network constants from a node service and remove dependency on tzscan (resolves #7, #63 and #64)

### LINKED ISSUE #7: FR: Remove dependency on tzscan
**Describe the solution you'd like**
All functionality of this application is dependent on tzscan.io. If tzscan is down, or inaccurate, or delayed, this tool carries the same issues.

**Describe alternatives you've considered**
This application should be using the local node's RPC-API to collect information.

**Additional context**
As of this posting, alpha.tzscan.io is incompatible with the new 003 protocol and is not showing accurate data. Because this application is dependent on data from tzscan, no testing using this app can happen at this time on alphanet.

### LINKED ISSUE #63: PRESERVED_CYCLES in Alphanet
The current network configuration suggests that NB_FREEZE_CYCLE is equal to two in the Alphanet network.
However, when executing the following command for preserved cycles:
`curl -s https://rpcalpha.tzbeta.net/chains/main/blocks/head/context/constants | jq .preserved_cycles`,
the obtained constant is equal to three.
I think the network configuration should be reviewed, and adjusted to be acquired dynamically from the node.

amzid:
- As described in the comments in issue #63, the network config constants should now be set from the local node access using the rpc interface of the client. In case the local node is not accessible, they can be acquired from a public node (tzbeta). If this is also not possible, then default values are set.
- The bug described in #64 is fixed through adding the lost rewards and fees to the calculation equation of total rewards
- The complete dependency on tzscan (described in issue #7) is avoided through the use of rpc calls on the running local node. This can be activated when setting `BOOL_USE_LOCAL_NODE` to `True` in the payment_producer script (should be configured externally for a long term use) in case a local node is accessible (which should be the case anyway for payment ability). 
The calculation of delegation status is performed through parsing the relevant snapshot block for a specific cycle. 
The total rewards were calculated through parsing the balance updates of the block where the rewards get unfrozen.

P.S.: The current version supports only the calculation of already unfrozen rewards. If the rewards are still frozen(in case of the use of a negative RELEASE_OVERRIDE), then the user is warned to wait until the rewards get unfrozen.

UPDATE: If the user opts for using a local node, then reward shares are computed twice (1.only from local rpc calls and 2. only from tzscan apis). If the obtained results do not match in any aspect, an exception is raised. In this way, results can be double checked for correctness before being forwarded to the payment customer.
If  `BOOL_USE_LOCAL_NODE` is set to `False`, the TRD would depend only on tzscan. I would suggest to keep the default value to `False` until this new feature is fully tested.

### DIFF
--- a/src/NetworkConfiguration.py
+++ b/src/NetworkConfiguration.py
@@ -1,6 +1,61 @@
-# todo: replace with rpc client rpc get /chains/main/blocks/head/context/constants
-network_config_map = {
-    'MAINNET': {'NAME': 'MAINNET', 'NB_FREEZE_CYCLE': 5, 'BLOCK_TIME_IN_SEC': 60, 'BLOCKS_PER_CYCLE': 4096},
-    'ALPHANET': {'NAME': 'ALPHANET', 'NB_FREEZE_CYCLE': 2, 'BLOCK_TIME_IN_SEC': 30, 'BLOCKS_PER_CYCLE': 2048},
-    'ZERONET': {'NAME': 'ZERONET', 'NB_FREEZE_CYCLE': 5, 'BLOCK_TIME_IN_SEC': 20, 'BLOCKS_PER_CYCLE': 128},
+from log_config import main_logger
+from util.rpc_utils import parse_json_response
+import requests
+
+logger = main_logger
+
+default_network_config_map = {
+    'MAINNET': {'NAME': 'MAINNET', 'NB_FREEZE_CYCLE': 5, 'BLOCK_TIME_IN_SEC': 60, 'BLOCKS_PER_CYCLE': 4096, 'BLOCKS_PER_ROLL_SNAPSHOT':256},
+    'ALPHANET': {'NAME': 'ALPHANET', 'NB_FREEZE_CYCLE': 3, 'BLOCK_TIME_IN_SEC': 30, 'BLOCKS_PER_CYCLE': 2048, 'BLOCKS_PER_ROLL_SNAPSHOT':256},
+    'ZERONET': {'NAME': 'ZERONET', 'NB_FREEZE_CYCLE': 5, 'BLOCK_TIME_IN_SEC': 20, 'BLOCKS_PER_CYCLE': 128, 'BLOCKS_PER_ROLL_SNAPSHOT':8},
 }
+
+CONSTANTS_COMM = "rpc get http://{}/chains/main/blocks/head/context/constants"
+URL = "https://{}.tzbeta.net/chains/main/blocks/head/context/constants"
+url_prefix = {"MAINNET" : "rpc", "ALPHANET" : "rpcalpha", "ZERONET" : "rpczero"}
+
+def init_network_config(network_name, config_client_manager, node_addr):
+    network_config_map = {}    
+    try:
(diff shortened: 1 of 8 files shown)


habanoz:
Thanks for your work. It is very valuable. I have some comments to make it better.

"if BOOL_USE_LOCAL_NODE" pattern can be changed with a factory pattern. Is there a reason for tzscan and RPC objects to coexist?

While deciding on the current implementation, the existence of a local node can be used. I hope the following statements can give an idea. 

class ProviderFactory:
@staticmethod
def getRewardApi()
    return RpcRewardApiImpl if local_node_present else TzScanRewardApiImpl
...
rewardApi = ProviderFactory.getRewardApiImpl()
...
reward_data = rewardApi.get_rewards_for_cycle_map()
...

I will spend more time to review the code. I may come back further propositions.

Thanks

habanoz:
Does "_config.yml" have a purpose?

amzid:
Thank you for your valuable feedback.
The idea about coexisting tzscan and RPC objects is to make testing easier in the short run at least. I think it is a good idea to compare both results and get an exception if anything does not match. Otherwise some little differences (or even big ones) may remain unnoticed. Until it is fully tested on all three networks, I would suggest keeping the coexistence of both instances.
For the long run, I think it is a matter of discussion. It can be a configuration option for the user to choose which interface to use, or even both for results double-checking since it is about money gains/losses.
However, if the community agrees that no matter what only one interface at a time should be used and decides to either leave that as a hard-coded setting or change it to a config parameter, I could easily change the code to fit the final decision (using the factory pattern).

I have overseen the file "_config.yml", I will try to exclude it from the PR since it is not used for now anyway. I would suggest in the same context to remove all files from the repo which do not have a purpose to make it easier to understand and more organized (e.g. businessConfiguration or network_configuration files)

I will be waiting for your further propositions.

habanoz:
I see why they need to coexist. However, heavy use of if blocks still does not look nice.  "api" package contains three interfaces: block_api, reward_api, reward_calculator_api. Those *_api interfaces are there almost from the beginning to enable and simplify future implementations. They are supposed to hide complexities of different implementations. 

Validation of rpc based results by comparing with tzscan based results is a good idea. We should keep it for a while.

We still need a validation mechanism such that results are compared with results from thirdparty services. Our calculation steps (steps after getting balance and reward data), either based on data provided by tzscan API or  rpc interface, may have bugs. If our calculation steps contains a bug, both calculation methods will produce the same in correct output. Tzscan webpage provides calculations at rewards tab, which can be used to validate against. This is future work.

amzid:
I wrote some code for validating calculations against rewards tab from TzScan, but the problem is that TzScan does not support the features of having a set of excluded delegators and a minimum delegation amount, which might lead to a divergence in the results due to a different configuation.
Since the calculation steps in the rpc and tzscan api are exactely the same, it is sufficient to compare the results given by the reward api as a first step.
The validation of calculation steps of both apis should be part of future work and discussed in another issue while considering the problem discussed above.

habanoz:
Yes, I agree this is future work.

habanoz:
I will merge this PR to pphases branch and run some tests. Do you mind if refactor it?

PR #68, opened 2019-04-04
Update Readme.md to include Python 3 install

Twente-Mining:
Added Python 3 install step


habanoz:
Thanks for your contribution. I merged it. However, I felt obliged to remove "curl" part. I do not feel it is safe to use third-party binaries.

PR #74, opened 2019-04-23
Add spaces to address: set values

nurikabe:
Without the space Python's `len()` comman assums that, for example, `tz1boot1pK9h2BVGXdyvfQSv8kd1LQM6H889:0.3` is a 40 character string.


habanoz:
You are right, without space file is not parsed correctly. It is not about len function but, yaml parser fail to parse.

Thanks for pointing out. Merging.

PR #78, opened 2019-05-22
Add support for p256 signatures from tz3 addresses

dordille:


habanoz:
Thanks for the PR. Can you please open an issue? The issue may give a brief description on why this is needed. Also if tz3 addresses are being used, validation logic needs to change, too.

dordille:
See #79 for further description of issue.

Regarding validation: Are you referring to the check that the signature is of the expected length [here](https://github.com/habanoz/tezos-reward-distributor/blob/7bb8e076e539be38c0e05588a9f4b4e495cf05bf/src/pay/batch_payer.py#L260)?

habanoz:
Actually I was referring to this line: https://github.com/habanoz/tezos-reward-distributor/blob/256886e913ed2e4a817d7c930a945a59b6beff16/src/util/address_validator.py#L12
however, on the second check I realized it is fine.

Thanks for #79 .  Merging PR.

PR #107, opened 2019-10-16
https://rpc.tulip.tools/ -> rpc.tulip.tools/

habanoz:


jdsika:
Just a fix for the rpc address. No need to mention it in the release notes.

Issue #104, opened 2019-10-14
tzscan close 


goodpen:
tzscan shutdown in this week. need new api

SemElVik:
+.
which explorer will we use now?  
let's see what the response from habanoz

goodpen:
maybe this tezos.id only this explorer show correct data

SemElVik:
total we have: 
`tezos.id` 
`tzstats.com` 

+ beta ver.
`https://mvp.tezblock.io` 
`https://arronax-beta.cryptonomic.tech/#/`

jdsika:
@amzid will Test the implementation using a local archive node. Start synchronizing now

MrZeke:
im running archive nodes.  Will i be able to still pay out 5 cycles in advance ?

utdrmac:
What about those of us not using archive mode?
Is there an option to use an external RPC endpoint instead of local node?

gizmokrang:
Maybe we can use this https://tzstats.com/docs/api/index.html#introduction

dawsy8888:
Hi ! what is the syntax to use -P rpc  ?  Also does the scrip use tezos-client for local rpc, and if so can I set the client config to use a remote rpc ?

gizmokrang:
https://gitlab.com/tezos-southeast-asia/tezos.id-doc/blob/dev/Api.md

habanoz:
We are waiting for @amzid to test on a local archive node. In order to activate rpc mode use with -P rpc option.

Not everybody may have an archive mode, for that case, there is public rpc mode which can be activated using -P prpc. The problem with prpc is we do not have a reliable public RPC service. If you know one, please share.

After switching to rpc/prpc mode it is not possible to pay in advance. Rpc implementation looks at the unfrozen rewards.

dawsy8888:
Does https://tzstats.com/docs/api/index.html#introduction help ?

tulpenhaendler:
Just wanted to say i run a public rpc endpoint here: 

https://rpc.tulip.tools/mainnet/chains/main/blocks/head

if that helps

For those of you running a archive node (@jdsika ) , there are snaphsots to speed up the process:

https://twitter.com/Blindripper85/status/1184155202762489857

without that syncing can take quite a while, make sure you have fast disks, ssd or nvme!

habanoz:
> https://rpc.tulip.tools/mainnet/chains/main/blocks/head

[...]

PR #108, opened 2019-10-16
back to mainnet.tezrpc.me

habanoz:


jdsika:
@habanoz will there be a list of possible endpoints collected from issue #104 ?

PR #122, opened 2019-10-26
Merge lrpc and prpc and make the node choice more flexible

amzid:
- The two classes PRpcRewardApiImpl and LRpcRewardApiImpl are merged in a single class RpcRewardApiImpl.
- The dependency on the tezos-client is removed for local rpc node calls, instead http requests are performed.
- The choice of a specific public node can be performed using e.g. -P rpc -A x.x.x.x:8732. The node should be an archive node.
- Edit readme in the part **how to run TRD** to add information about how to specify the data provider. This part will be moved to the sphinx documentation when the other branch will be merged.


jdsika:
Are the formerly disabled tzscan unit test enabled again?

PR #132, opened 2019-11-10
Fix tzstats and destination map issues

### LINKED ISSUE #120: destination_map does not redirect payments
**Describe the bug**
I specified a payout redirect in the .yaml configuration file under rules_map. Unfortunately, the reward was not redirected and was not displayed in the calculation CSV. It is possible that destination_map does not work? 

**To Reproduce**
Our rules_map looks as follows:
rules_map:
  KT-Contract-1: TOF
  KT-Contract-2: TOF
  TZ-Address-1: TOF
  TZ-Address-2: TOF
  **KT-Contract-3: TZ-Address-3**
  mindelegation: TOB

_KT-Contract-3 and TZ-Address are also defined under supporters_set_

**Expected behavior**
I did expect that the Payouts will be redirected from KT-Contract-3 to TZ-Address-3. 

**Screenshots**
When running:
```
  ],
  "__destination_map": {
   "KT-Contract-3": "TZ-Address-3"
  }
```

**Environment**
 - Source Code Build (c4c5c9f886c02785d1dc886c1e9ebd59d4dd3448)
 - Archive-Node on Mainnet-Staging
 - python3 src/main.py -C 156 -D -P rpc -M 3 --verbose

Edit: I also checked the case where KT-Contract-3 and TZ-Address are not included in the supporters_set. Same issue.

### LINKED ISSUE #128: -R [-6,-11] is not working properly
**Describe the bug**
With the use of an rpc and prpc node the range [-1,-11] must be deactivated and it must be noted, that the range [-1,-5] can be used with tzstats.

The range [-7, -11] "payment of future rewards" does not work and must be turned off.
-R -6 must be denoted as "currently available rewards in already achieved in this active cycle"
and also be deactivated, since tzstats only displays the actual rewards and no future rewards.
An issue in the GitHub of tzstats must be opened that requests "future possible rewards" if we want it to be processes in TRD.
For now at range [-6,-11] show a message in the log that these are not yet achieved future rewards that are not yet supported by tzstats and the local node.

**To Reproduce**
At cycle 165
python3 src/main.py -C 165 -D -M 3 -R -6
python3 src/main.py -C 170 -D -M 3 -R -11

amzid:
* Fix tzstats issues: As noted in the last PR, tzstats was in testing period and was therefore not set as default. We noticed that the staking balance given by tzstats should be retrieved from the snapshot call and not from the income call, otherwise (differently from tzscan) the staking balance of the payout cycle and not of the snapshot cycle will be taken as reference.
Furthermore, we added the possibility to get reward data of upcoming cycles based on the expected income (resolves #128).
HOWEVER, since tzstats does not allow commercial use only under previous agreement and that could be taken for granted, we will keep on not using tzstats as default. PRPC will be still the default, using tzstats should be explicitly set through the -P tzstats flag, and under personal guarantee in case of commercial use.
* destination map issues (resolves #120)

### DIFF
--- a/src/calc/phased_payment_calculator.py
+++ b/src/calc/phased_payment_calculator.py
@@ -86,6 +86,7 @@ class PhasedPaymentCalculator:
         phase4 = CalculatePhase4(self.founders_map, self.owners_map)
         rwrd_logs, total_rwrd_amnt = phase4.calculate(rwrd_logs, total_rwrd_amnt)
 
+
         # calculate amounts
         phase_last = CalculatePhaseFinal()
         rwrd_logs, total_rwrd_amnt = phase_last.calculate(rwrd_logs, total_rwrd_amnt)
--- a/src/pay/batch_payer.py
+++ b/src/pay/batch_payer.py
@@ -248,7 +248,7 @@ class BatchPayer():
 
             op_counter.inc()
 
-            content = CONTENT.replace("%SOURCE%", self.source).replace("%DESTINATION%", payment_item.address) \
+            content = CONTENT.replace("%SOURCE%", self.source).replace("%DESTINATION%", payment_item.paymentaddress) \
                 .replace("%AMOUNT%", str(pymnt_amnt)).replace("%COUNTER%", str(op_counter.get())) \
                 .replace("%fee%", self.default_fee).replace("%gas_limit%", self.gas_limit).replace("%storage_limit%", self.storage_limit)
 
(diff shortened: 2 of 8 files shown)


jdsika:
* Logging info missing which tells the user that he is using "future payouts" where the rewards are estimated and not earned yet
* Logging message which tells the user that he is paying out frozen rewards which intheory can still be slashed
* Note in the csv is missing which informs about the redirect of the payment

amzid:
I will edit the reporting and logging process as suggested by @jdsika and commit that seperately.

Issue #145, opened 2019-12-15
Issues with Batch Payments


dansan566:
Describe the bug
Here is a collection of bugs concerning batch payments that I collected together with another Tezos Baker (Bake'n'Rolls).

To Reproduce

1. Issue: Patch payments cannot be executed if an address has been removed from the state tree. Quick-Fix: Exclude address and execute separately (incl. payment of burn fee) 
2. _Issue This time, 1 batch failed because of "Transaction Error proto.005-PsBabyM1.contract..empty_transaction" for this guy: https://tzstats.com/tz1cKfV8LNBGQ6DSnzR63j8Rq5g43AXbEJsD
The payout amount was 0.000000 XTZ and his balance was 4XTZ at the moment of payment.
Since 0.000000 XTZ payments are passing to the KT1 addresses, it seems that the issue is in the payout of 0 XTZ to the tz1 address.
Probably the resolution here is to avoid 0.000000 XTZ payments at all because they are no sense :)_

jdsika:
@dansan566 can you please verify the fix and close the issue if you are ok with the solution?

jdsika:
I think it is solved according to the conversations

PR #158, opened 2019-12-18
Remove trivial, alter minimum payout calculation

### LINKED ISSUE #141: Changing minimum/trivial payout formula
We are getting lots of small delegations which translate to small reward payouts. I've noticed this log line that says "X non-trivial payments, total XXX.XXX". This made me dig into the code where I see that a payment is only considered payable if it is greater than 2e3 (2000) mutez (or 0.002 XTZ). This will certainly raise some comments about people not being rewarded.

Shouldn't the formula be: `if (reward - txnFee > 0.0){send_reward();}` ?
Obviously, if the baker is absorbing the txnFee themselves, then this translates to:
  `if (reward > 0.0){send();}`

### LINKED ISSUE #147: Trivial/Not-Trival are backwards
**Describe the bug**
The word "trivial" is defined as "of very little importance or value; insignificant:"
Source: https://www.dictionary.com/browse/trivial?s=t
In the code, there are references to "non-trivial" and "trivial" payments. In the code, these are backwards; trivial references payments which will be made (which is not correct), and non-trivial payments which will not be made because they are too small.

This backwards definitions causes confusion while reading the code.

I will submit a PR to fix this later today.

**To Reproduce**
See code in src/pay/batch_payer.py

**Expected behavior**
N/A

**Environment**
All environments

utdrmac:
Fixes #147 
Fixes #141

### DIFF
--- a/src/pay/batch_payer.py
+++ b/src/pay/batch_payer.py
@@ -32,8 +32,7 @@ COMM_INJECT = " rpc post http://%NODE%/injection/operation with '\"%OPERATION_HA
 COMM_WAIT = " wait for %OPERATION% to be included --confirmations {}".format(CONFIRMATIONS)
 
 FEE_INI = 'fee.ini'
-DUMMY_FEE = 1000
-
+MUTEZ = 1e6
 
 class BatchPayer():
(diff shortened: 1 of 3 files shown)


utdrmac:
Fixes #145

jdsika:
@amzid please review

Thank you @utdrmac for the contribution!

PR #164, opened 2019-12-27
Change default public rpc node

amzid:
The default nodes are now as following:
* "https://api.tezos.org.ua/" for mainnet
* "https://tezos-dev.cryptonomic-infra.tech/" for test-net (alphanet, babylonnet etc)
* "https://rpczero.tzbeta.net/" for zeronet


jdsika:
@amzid should this be merged now?

PR #168, opened 2019-12-30
Hostname, port to client binary call; fix prior Athens issues

### LINKED ISSUE #156: Failed to acquire the protocol version from the node
I have a node running on my LAN and I mapped it as "tezos-node-lan".

My configuration is:
```
MainThread - Arguments Configuration = {
 "initial_cycle": 170,
 "run_mode": 1,
 "release_override": 0,
 "payment_offset": 0,
 "network": "MAINNET",
 "node_addr": "tezos-node-lan:8732",
 "reward_data_provider": "rpc",
 "node_addr_public": "https://mainnet.tezrpc.me",
 "reports_base": "/tezos-reward-distributor/pymnt/reports",
 "config_dir": "/tezos-reward-distributor/pymnt/cfg",
 "dry_run": true,
 "dry_run_no_consumers": false,
 "executable_dirs": "/usr/local/bin",
 "docker": false,
 "background_service": false,
 "do_not_publish_stats": false,
 "verbose": true
}
```

I get this error:
```
Failed to acquire the protocol version from the node
Rpc request failed:
- meth: GET
- uri: http://localhost:8732/chains/main/blocks/head/protocols
- error: Unable to connect to the node: "Connection refused"
 ```

It should be `http://tezos-node-lan:8732/chains/main/blocks/head/protocols`.

I can't find in NetworkConfiguration.py anything related to this RPC GET command.

### PR DESCRIPTION
**Pull request by utdrmac**
Fixes #156

### DIFF
--- a/src/rpc/rpc_reward_api.py
+++ b/src/rpc/rpc_reward_api.py
@@ -14,7 +14,7 @@ class RpcRewardApiImpl(RewardApi):
     COMM_DELEGATES = "{}/chains/main/blocks/{}/context/delegates/{}"
     COMM_BLOCK = "{}/chains/main/blocks/{}"
     COMM_SNAPSHOT = COMM_BLOCK + "/context/raw/json/cycle/{}/roll_snapshot"
-    COMM_DELEGATE_BALANCE = "{}/chains/main/blocks/{}/context/contracts/{}"
+    COMM_DELEGATE_BALANCE = "{}/chains/main/blocks/{}/context/contracts/{}/balance"
 
     def __init__(self, nw, baking_address, node_url, verbose=True):
         super(RpcRewardApiImpl, self).__init__()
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by utdrmac** [comment]
Ah, I didn't see all those test scripts that need this change too.

**Comment by utdrmac** [comment]
```
$ python3 -m unittest discover -s src
..............
----------------------------------------------------------------------
Ran 14 tests in 166.562s

OK
```

**Comment by jdsika** [comment]
I have to wait for @amzid to give a go here before I merge. Is there trouble with the use of the tezos-client? We should replace it by CURL?

**Comment by utdrmac** [comment]
@jdsika We should attempt to remove all usage of `tezos-client`. It is [incredibly inefficient](https://gitlab.com/tezos/tezos/issues/587#note_264393648). Aside from `list addresses`, I don't think there are any other client RPC calls that cannot be turned into simple python.requests()

**Comment by utdrmac** [comment]
@jdsika It was discovered that if you attempt to use the client like this (how TRD does it):
`tezos-client rpc get http://some.provider.com/chains/mainnet/blocks/header`
The client ignores/strips 'http://some.provider.com' from the URL and always attempts to access localhost. There are specific -client parameters (-A and -P) to specify alternative hosts.

File: src/rpc/rpc_reward_api.py [hunk 1]
```diff
@@ -138,32 +139,28 @@ def __get_delegators_and_delgators_balance(self, cycle, current_level):
...
-                    try:
-                        response = self.do_rpc_request(request, time_out=5)
-                    except:
-                        logger.debug("Fetching delegator info failed {}, will retry", delegator)
+        # loop over found delegators
+        for idx, delegator in enumerate(delegators_addresses):
+            request = self.COMM_DELEGATE_BALANCE.format(self.node_url, hash_snapshot_block, delegator)
 
-                delegators[delegator] = int(response["balance"])
+            try:
```
amzid: Shouldn't the loop around this fetch operation be kept? (in case a large number of delegators, one or more rpc call(s) may fail, if this fetch is not retried, the balance will be considered as zero (https://github.com/habanoz/tezos-reward-distributor/commit/4494abf7e2f874d18f57757560d21f10dbb953bc)

utdrmac: Yes, there should be some retry logic. But what was there previously made RPC calls twice for no reason. If you look here, https://github.com/habanoz/tezos-reward-distributor/blob/d1cfad3d1c7738a999c381ca25afd74cd04aaf8d/src/rpc/rpc_reward_api.py#L148, you see that request is set, then it sleeps, then sets to none, then does it again. That's a complete waste of a request.

I'll add the loop back in but without calling RPC twice

amzid:
Thank you for the great contribution and for solving the above mentioned issues. This looks good to me, and works for preliminary tests in mainnet and babylonnet (testnet).

PR #183, opened 2020-01-20
Update sphinx documentation

amzid:
The plan is to move the whole wiki parts to a more organized and accessible sphinx documentation. Some parts of the documentation still need to be reviewed and updated to the current state of code. We are also planning to move all used constants to a single file (Constants.py) to make adapting code to protocol changes more flexible (e.g. default ports or currently hardcoded storage limits)

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1,24 +1,5 @@
-
 How to configure Tezos Reward Distributor?
-=====================================================
-
-Baker Configuration:
-------------------------
-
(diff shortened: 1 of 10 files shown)


amzid:
Yes, it builds locally. I would suggest to merge it to be able to see if it gets deployed correctly and to be able to see the current stand live under https://habanoz.github.io/tezos-reward-distributor/. Afterwards, we can create an issue to discuss the current content and identify the parts that need to be reworked.

PR #184, opened 2020-01-20
Added rules_map, min_delegation_amt and re-ordered.

dansan566:


docs/configuration.rst
```diff
@@ -48,38 +68,34 @@ Available configuration parameters are:
...
   needed to be added.
 
-**specials_map**
-  A dictionary of PKH and fee (decimal in the range [0-100] )
-  pairs. This dictionary can be used to set special service
-  fee values for desired delegators.
-
+**min_delegation_amt**
+ A minimum delegation amount can be set here. If this value is set 
+ to 10, 10 XTZ are required as minimum.
```
jdsika: Can you specify what happens to the payout amount if it is under the minimum and how it is documented?

docs/configuration.rst
```diff
@@ -48,38 +68,34 @@ Available configuration parameters are:
...
-  or alias of implicit or originated account is accepted. For
-  more information on the payment address configuration please
-  refer to the next section.
+**rules_map**
+  The rules_map is needed to redirect payments. Constants are TOF = To Founders Balance, 
+  TOB = To Bakers Balance and mindelegation.
+  
+  Example: 
+  rule_map:
+  PKH: TOF (redirects payment from PKH to TOF)
```
jdsika: We need one section with all abbreviations

Issue #192, opened 2020-01-26
Missing new parameters causes errors


utdrmac:
If a user does not update their yaml config file to include the new parameters introduced by #181 they receive the following error message:

```
Traceback (most recent call last):
  File "src/main.py", line 207, in <module>
    main(args)
  File "src/main.py", line 157, in main
    reactivate_zeroed=cfg.get_reactivate_zeroed(),
  File "/home/XX/code/tezos-reward-distributor/src/model/baking_conf.py", line 75, in get_reactivate_zeroed
    return self.get_attribute(REACTIVATE_ZEROED)
  File "/home/XX/code/tezos-reward-distributor/src/model/baking_conf.py", line 45, in get_attribute
    raise Exception("Attribute {} not found in application configuration.".format(attr))
Exception: Attribute reactivate_zeroed not found in application configuration.
```

Will submit PR soon that aborts startup if there are missing parameters in config.

PR #193, opened 2020-01-26
Exit on failure to parse missing config parameters 

utdrmac:
- Treat new parameters as having no default, forcing users to pick their own setting


utdrmac:
Fixes #192

jdsika:
@amzid for review. I think it is kind of mandatory to check in the config parser if all parameters are either given in the config file or the standard values are set by the parser and communicated to the user.

PR #202, opened 2020-02-13
Skip KT1 accounts when checking 0 balance

### LINKED ISSUE #201: Empty KT1 do not require reactivation
**Describe the bug**
Example: KT1VWyRSC1qLoZaytv6Z9UUftcD1WYnSDByg has 0 balance. If it were a tz1, you would be required to pay the burn fee to reactivate. But this is not the case for KT1 accounts. KT1 accounts to not need to be "reactivated" if their balance is 0

**Misc**
Special thanks to Dmitry (BakingBad) for discovering this.

utdrmac:
Resolves #201


jdsika:
@amzid for review

PR #205, opened 2020-02-25
"use_ssl" -> "use.ssl"

gracenoah:
Reference:
https://github.com/habanoz/tezos-reward-distributor/blob/ca49256c078278df8ff94f06cc4f4e1c6e224eed/src/emails/email_manager.py#L16

### DIFF
--- a/examples/email.ini
+++ b/examples/email.ini
@@ -5,4 +5,4 @@ smtp.host =
 smtp.port = 
 sender = 
 recipients = 
-use_ssl=
+use.ssl=


jdsika:
@utdrmac please check - I will release the next two days. @amzid will do a branch for Carthage compatibility. After activation we will merge and release a tag for carthage compatibility with a warning that it needs to be tested and feedback is needed.

utdrmac:
Yep, good catch @gracenoah

Issue #188, opened 2020-01-23
TF Grant Submission


utdrmac:
Hello @habanoz , @amzid , @jdsika ,
I think it's fair to say that TRD is one of the, if not the, most used rewards payments system for Tezos. This project/tool should have TF grant funding behind it. If TF can give a grant to a for-profit company, they can give one here.
I'll be submitting the grant request later this week. I do not plan to request the max amount; probably somewhere around 40% of the max. Any funds that I acquire from the grant I plan on distributing to TRD code contributors and documentation writers as we all deserve compensation for our time in making this tool what it is today.

jdsika:
If you submit a grant then only for new features of the tool. 
* Support for the protocol upgrades in 2020
* Extension of the documentation
* Monitoring for the run and forget mode

dansan566:
I also think it's a good idea to ask for a grant. Especially @habanoz , @amzid and you have already spent so much time on this that this is only justified that further workload will be compensated in the future. 

However, I have to agree with @jdsika that a grant may only be used for future development and features. It would also provide an incentive to do so. I would like to hear the opinion of @habanoz - as far as I am informed a request was rejected at that time.

habanoz:
I am sharing my past conversation with Tezos Foundation. It was not a private chat thus I believe it is appropriate to share.

habanoz:
Dear Huseyin,

Thank you for applying for a grant from the Tezos Foundation. As you know, we receive a significant number of strong proposals. After conducting a full review of your project, the Foundation has decided not to fund your project at this time.

We sincerely value your efforts as a member of the community and hope that you will continue to be involved in the Tezos project. We appreciate your patience as your proposal was evaluated, and wish you the best of luck going forward.

All the best,
Charlie Wiser

habanoz:
Dear Charlie, 

Thanks for your response. I will continue being a member of the tezos community.

As an ICO investor an active community member I need to understand why the proposal is rejected. I need to know that because I need to see that the foundation is doing wisely while vesting the money we devoted to project success. I also need to see that the foundation is doing wisely while appreciating the effort put to project success. 

I know that my proposal fills a gap in tezos ecosystem. It is not a concept but a working product. I receive emails from bakers presenting their pleasure and gratitude. I put a lot of effort to build it. With respect, I doubt that proposals are evaluated decently. I can see that the video I uploaded is not watched at all. 

[...]

PR #227, opened 2020-05-26
Parse node urls to handle potential tls encryption (resolves #226)

### LINKED ISSUE #226: Bugs when using secured nodes (https)
**Describe the bug**
I tried to configure a secured node (e.g. https://mainnet.tezrpc.me) to be used by the TRD. The client within the TRD is however set to use only http, but no https nodes. 

**To Reproduce**
E.g. start the trd using these arguments `-V -N ALPHANET -A rpcalpha.tzbeta.net:443 -P rpc -C 174 -M 3` or any node using the https protocol.

The following errors are obtained:

```
Command failed, error is |Warning:
  Failed to acquire the protocol version from the node
  Rpc request failed:
     - meth: GET
     - uri: http://rpcalpha.tzbeta.net:443/chains/main/blocks/head/protocols
     - error: Oups! It looks like we forged an invalid HTTP request.
```

```
  File "/usr/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/home/amzid/02_StakeNOW/tezos-reward-distributor/src/pay/payment_producer.py", line 107, in run
    crrnt_cycle = self.block_api.get_current_cycle()
  File "/home/amzid/02_StakeNOW/tezos-reward-distributor/src/api/block_api.py", line 18, in get_current_cycle
    return self.level_to_cycle(self.get_current_level())
  File "/home/amzid/02_StakeNOW/tezos-reward-distributor/src/rpc/rpc_block_api.py", line 20, in get_current_level
    head = response.json()
  File "/home/amzid/.local/lib/python3.7/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3/dist-packages/simplejson/__init__.py", line 518, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3/dist-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/usr/lib/python3/dist-packages/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
  File "/usr/lib/python3/dist-packages/simplejson/scanner.py", line 79, in scan_once
    return _scan_once(string, idx)
  File "/usr/lib/python3/dist-packages/simplejson/scanner.py", line 70, in _scan_once
    raise JSONDecodeError(errmsg, string, idx)
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

```

amzid:
This PR resolves the issue #226. The following steps were performed:

* **Bug Analysis**: The obtained errors in issue #226 were analyzed. As described in the issue, they were caused by two main reasons:
    * The client fails to get any information because of a failing -S argument. This causes any request from the client to the tls-encrypted node to fail.
    * The rpc block api fails because of the hardcoded http protocol. This causes the second error described in the issue.

* **Solution**: Add the possibility to include the protocol prefix in the node url. If https is used, the node is accessed using tls encryption.

* **Implementation**: 
    * The node url given using the `-A` can now contain a `https` prefix. If it is the case, this prefix is parsed and an additional `-S` argument is given to the tezos client for all rpc requests.
    * The hardcoded protocol prefix in the rpc block api is replaced by the given protocol prefix.

* **Performed tests**: the performed tests include the use of multiple combinations of nodes. The use of tls-encrypted public nodes (e.g. https://rpcalpha.tzbeta.net:443) for acquiring the network protocol and injecting operations is now possible.

* **Documentation**: The help section is adjusted to show the possibility of configuring the tls encryption in the node url.

**Work effort**: One half day (around 4 hours).


jdsika:
Please add a PR template and the section about your work effort

jdsika:
@amzid can we ensure that encrypted communication to nodes is done by default and http only if it turns out encryption is not possible? I think it is a security issue to have unencrypted communication?

@utdrmac what do you think about the PR style? Do you think this would be sufficient with regard to #188 ? Can you review the PR?

src/api/provider_factory.py
```diff
@@ -25,6 +25,8 @@ def newRewardApi(self, network_config, baking_address, node_url, node_url_public
 
     def newBlockApi(self, network_config, node_url):
         if self.provider == 'rpc' or self.provider == 'prpc':
+            if node_url.find("http") == -1:
+                node_url = 'http://' + node_url
```
jdsika, line 29: Can't it be https:// by default?

src/cli/simple_client_manager.py
```diff
@@ -9,17 +9,26 @@ def __init__(self, client_path, node_addr, verbose=None) -> None:
         self.client_path = client_path
         self.cmd_manager = CommandManager(verbose)
         self.node_hostname = "127.0.0.1"
-        self.node_port  =  8732
+        self.node_port = 8732
+        self.tls_on = False
```
jdsika, line 13: should this then be true?

src/launch_common.py
```diff
@@ -94,7 +94,7 @@ def add_argument_network(parser):
 
 
 def add_argument_node_addr(parser):
-    parser.add_argument("-A", "--node_addr", help="Node (host:port pair). Default is 127.0.0.1:8732. This is the main Tezos node used by the client for rpc queries and operation injections.", default='127.0.0.1:8732')
+    parser.add_argument("-A", "--node_addr", help="Node (host:port pair) potentially with protocol prefix especially if tls encryption is used. Default is 127.0.0.1:8732. This is the main Tezos node used by the client for rpc queries and operation injections.", default='127.0.0.1:8732')
```
jdsika, line 97: do we do line breaks for too long lines? Isn't PIP8 or something automatically checked in the CI?

utdrmac:
There should be a test somehow for https because `tezos-node` doesn't use https by default, so might this may break some installs?

PR template is good. When/where does the submitter put their PKH for compensation?

amzid:
I have the same opinion as @utdrmac. I think we should keep http by default since it is also the default of `tezos-node`. It might be confusing for the users if we change the default to https.

src/launch_common.py
utdrmac: Should we make the default be "http://127.0.0.1:8732"?

jdsika (approved):

jdsika:
@utdrmac please approve and I will merge afterwards with the right commit message

PR #229, opened 2020-05-28
[WIP] Create docker file for trd

### LINKED ISSUE #228: Create a docker image
I want to base our bakery's infra on docker, and would love to know what you think about that and what can we do? I started to work on it, my current solution is hacky, but I'll open a PR anyway so you can view my work.

chiptus:
close #228


chiptus:
I'm making some progress. This is my first dockerfile, it's working right now with some hacks. need to mount tezos-client into `/tezos-client` and mount its config into `/trd/tezos` (also trd config and reports needs to be mounted)

I'm open to hearing suggestions. Currently trying to test using a teztool sandbox, blocked because the sandbox is created always with archive mode set to full

utdrmac:
This probably needs support for the various -M options as well as being able to start from a specific cycle (-C)

chiptus:
it supports any flag that you'll pass it

utdrmac:
I'm looking at your ENTRYPOINT and I don't see where you are including external parameters. I'm not a docker expert in any way, but it was my understanding that you have to add some type of `@extra` parameter so that additional parameters on the CLI are passed to the entrypoint script.

chiptus:
you run it like this `docker run trd --extra-parameter` the parameters are
passed to the entry point

amzid:
Thank you for the effort. Is it possible to provide the motivation behind your PR (use case) and a step-by-step guide on how to get it run?

chiptus:
The use case would be to add it to our bakery stack for easy deployment and
management. using a tool like portainer or other docker/kube management
tools is much easier when everything is dockerized.

actually this is really a PoC, I really don't like the run method, for now
it's something like this:

```
docker build -t trd .
 docker run -v `pwd`/test:/trd/config -v `pwd`/test-reports:/trd/reports -v
`which tezos-client`:/tezos-client -v /home/chaim/.tezos-client:/trd/tezos
-it --rm -network host trd:latest
```

needs to have tezos-client installed and your config inside
tezos-config-docker.

Again, this is only WIP.

I want it just to be `docker run --rm trd:latest` maybe with configuration
of the bakery and how to contact the node

utdrmac:
As you've stated this is PoC/WIP, we can just leave this PR open until you are satisfied with what you have contributed.
Yes, getting the tezos-client working will be extremely difficult.
To simplify things, I would stick with a single volume `/trd` instead of multiples. I don't see the reason for multiple volumes contained in the same path.

jdsika:
@utdrmac I think we had in mind no remove the dependency on the tezos-client and use curl anyway?

chiptus:
That would be great

chiptus:
And sorry, I never found the time to work on this

utdrmac:
Yes, we should try to stop using the tezos-client directly. It would help with situations like Docker, K8S, and Kiln.

jdsika:
@chiptus the dependency from tezos-client has been removed in #294 I think you can return to this work item now

jdsika:
@habanoz maybe you can take this over?

jdsika:
@amzid status for merge today?

jdsika:
Closing because it has been solved in #442

Issue #204, opened 2020-02-24
Compensate delegators for missed endorsements/bakings


chiptus:
**Is your feature request related to a problem? Please describe.**
As a baker, I want to compensate my delegators for missed blocks.

**Describe the solution you'd like**
I think that a boolean config would be great to indicate if I'm paying for missed blocks. We can take it another step and supply percent, to indicate how much of the missed payment I want to pay, e.g for each missed endorsement I would pay 1tz instead of 2

**Describe alternatives you've considered**
I'm considering writing an external script to do that, but then it will be left out of the CSV files. A single solution for the whole process is what I'm looking for.

utdrmac:
@chiptus The problem with implementing such a feature is that it would require TRD to be aware of all baking/endorsing rights for a particular baker.
`/chains/main/blocks/head/helpers/endorsing_rights` and `/baking_rights` are very CPU intensive RPC calls and most public RPC providers block these endpoints. I believe that this would force anyone wanting this payout methodology to have a local archive node.

This could be implemented as a basic option to TRD in how you pay as a baker: A) We pay actual rewards, or B) We pay idealized rewards.  A is how TRD functions today. B could be implemented by making the above RPC calls and doing all rewards calculations based off network constants (since the reward is now variable in Carthage) and then divided up as normal.

I'll put this in our list of features for consideration.

Issue #69, opened 2019-04-06
delegators those overdelegate


habanoz:
Currently, rewards are shared among all delegators. If the delegate is over-delegated this is not desirable for delegators. 

It is necessary to implement a mechanism to avoid paying to delegators those over-delegate.

jdsika:
Enforcing proper behaviour according to overdelegation is important. Over-delegation hurts the baker as it ruins his statistics for the important rankings. Partial rewards from the overdelegation should be payed to the baker. We have to check which was the last address that did the delegation and caused the over-delegation. A problem: This can not only happen with a new delegation but also with stocking up on an already delegating address by increasing the funds. The behaviour needs to be clearly documented.

jdsika:
I think we should as well include the possibility to define this payment matrix:

![Bildschirmfoto 2019-09-10 um 15 26 50](https://user-images.githubusercontent.com/11943343/64621537-23519a00-d3e6-11e9-8b13-a6eb26f793c5.png)

habanoz:
Baking config (yml) file can be used to select payment matrix. Do you think this is essentials for bakers?

habanoz:
> We have to check which was the last address that did the delegation and caused the over-delegation. A problem: This can not only happen with a new delegation but also with stocking up on an already delegating address by increasing the funds. The behaviour needs to be clearly documented.

It is not easy to keep track of delegations. Tzscan creates a list of delegations. This can be used for tzscan backend. What about rpc backends?

chiptus:
>We have to check which was the last address that did the delegation and caused the over-delegation. A problem: This can not only happen with a new delegation but also with stocking up on an already delegating address by increasing the funds. The behavior needs to be clearly documented.

I wonder how can I find this by hand, like if I go to an explorer will I be able to see this? As I see it, for now, if it's a new delegation, the easiest would be to give the new delegation 0% from its rewards.

SemElVik:
of course you can see.
for example here: `https://tzkt.io/tz1......................../delegators`

jdsika:
This issue will resolve itself with Tenderbake I think. Check new specification

PR #232, opened 2020-06-11
TzKT reward & block provider

m-kus:
This is an implementation of the `BlockApi` and `RewardApi` interfaces for [TzKT API](https://api.tzkt.io/).  
All methods and dry-run are covered with tests, although not suited for running in travis pipeline (takes about 10min), probably one can use dumps from Rpc/Tzstats providers instead of querying them each time.
  
Please note, that there can be differences in final payout amounts due to how different providers handle different rewards:  
* Tzstats provider does not handle accusation rewards/losses
* Both RPC and Tzstats calculates incorrect staking balances for cycles with snapshot@16 (see https://medium.com/@_MisterWalker_/we-all-were-wrong-baking-bad-and-most-bakers-were-using-wrong-data-to-calculate-staking-rewards-a8c26f5ec62b)
  
Also, TzKT allows to distinct between cases when rewards are missed due to overdelegation and other reasons. Also, you can track the order of delegations and determine which particular delegator is not to be paid. This can probably help with #69 and #204 
  
Generally, it would be great if TRD supports the [Bakers Registry](https://hackmd.io/DZyJU5HSThmGX8ER7tprXg?both) config format in order to unlock the whole potential of TzKT/Tzstats reward splits.  

Would love to answer any questions or add more tests if needed :)


m-kus:
For some reason Travis build machine cannot reach our API host :( Anyways, I've disabled those tests on travis, still they can be run locally.  
Also added more tests for edge cases and file cache for `rpc` / `tzstats` providers output, now it takes only few seconds to run.  
  
One more thing, I've added an extra cmdline argument for passing custom API endpoint (for local instances of TzKT/Tzstats), it's None by default so should not break anything.  

Please review and let me know if one need to fix/add/remove anything.

jdsika:
@amzid can you help with Travis ci?
Do you need some kind of whitelisting to reach the server?

amzid:
> @amzid can you help with Travis ci?
> Do you need some kind of whitelisting to reach the server?

I am also not sure why this happens. I would suggest contacting the support of Travis CI and sending the log of the unsuccessful build of commit 9f7a3e2.

Groxan:
Any updates on this PR?

amzid:
For me the tests are working locally as well. I contacted the travis CI support, if we do not receive a response within the next week, we might need to deactivate the tests on the CI.

amzid:
The Travis CI support suggests trying the following work around https://stackoverflow.com/a/33184416 and is ready to follow up on this issue.

m-kus:
I surely can mock TzKT requests, but it would make less sense since it doesn't guarantee TzKT won't respond differently in the future. I think disabling theese tests running on Travis is a preferable option, let me know which of two options works best for you.

amzid:
In this case, I think the implementation overhead is not worth it since it is not guaranteed to be stable. Since the tests can be anyway performed locally, I personally also opt for the option of disabling them in Travis.

src/tzkt/tzkt_block_api.py
```diff
@@ -0,0 +1,24 @@
+from tzkt.tzkt_api import TzKTApi, TzKTApiError
+from api.block_api import BlockApi
+
+
+class TzKTBlockApiImpl(BlockApi):
+
+    def __init__(self, nw, verbose=False, base_url=None):
+        super(TzKTBlockApiImpl, self).__init__(nw)
+        if base_url is None:
+            self.api = TzKTApi.from_network(nw['NAME'].lower(), verbose=verbose)
```
amzid, line 10: In the current implementation of the TRD, we fix the development network name to Alphanet, but it is meant to be the current active test network. I think tzkt expects the name of the active test network and not alphanet, right?

amzid (approved):
This PR is ready to be merged from my side. In future PRs, we will work on a strategy to isolate tests from the core code.
Was this work already paid in the context of a grant/company income, or should the work effort be compensated by the TRD grant?

m-kus:
> This PR is ready to be merged from my side. In future PRs, we will work on a strategy to isolate tests from the core code.
> Was this work already paid in the context of a grant/company income, or should the work effort be compensated by the TRD grant?

Great! No additional compensation is needed.

jdsika:
Are the tests now running locally with a commit hook?

Issue #116, opened 2019-10-21
insufficient funds


goodpen:
if you dont have enough tezos on balance to pay you received hash https://tzstats.com/operation/ooujFHaXQecd24mEa6Eikk8hMz6eqoLydxNnYBUhmpj3C4yNdX7

but operation faild, maybe TDR must warring about this (dont enough found to pay)

jdsika:
@amzid I would classify it a bug if the funds are not checked before forging an operation - what do you think?

habanoz:
Run_operation is supposed to fail if funds are not enough. 

Are you able to replicate the situation?

goodpen:
> Run_operation is supposed to fail if funds are not enough.
> 
> Are you able to replicate the situation?
> 
> Do you have logs ?

in next cycle i replicate the situation and share log

goodpen:
payment wallet not enough xtz but give me hash ongMAfRM4FvsCgzbpSRAyu4N66xBqaDEQ4zLtUrsLirUJov4vEt

log in attach 
[log.txt](https://github.com/habanoz/tezos-reward-distributor/files/3775572/log.txt)

amzid:
This issue was reformulated in #220 and is solved by #251. Feel free to reopen the issue if needed.

Issue #220, opened 2020-05-07
Check for sufficient funds before payment


jdsika:
**Is your feature request related to a problem? Please describe.**
Most bakers use a "payment address" which is a "non ledger" address for automated payments using tezos-signer. You always have to check if enough funds are on the address used for the payment.

**Describe the solution you'd like**
I would like to check if the address has sufficient funds to do the payment. If not put the payment in a "to do queue", show a warning message and if configured send an email.

**Additional context**
It would be nice to configure a "minimum amout" on the payment address and send out a reminder email if the threshold is breached.

jdsika:
I think this would also solve the problem of getting the wrong impression of paid rewards by receiving the rewards email notification in service mode?
If there are not enough funds it should gracefully abort and in service mode send a mail with the notice "insufficient funds" and retry the payment in X cycles

utdrmac:
#245 Could be improved to specifically catch the "insufficient funds" error and display better logging message.

jdsika:
@amzid is on it

jdsika:
This is implemented with #251

Issue #149, opened 2019-12-15
CycleNumber.csv moved to Done after failed payment


Xa52c:
Hot wallet contains no funds.
Launch TRD as a service.
It runs to make a payment for cycle number X.
No payment is made (hot wallet empty).
File X.csv gets put into Done.
Transfer funds to hot wallet.
TRD does not retry failed payment.

jdsika:
Thank you reporting. We will look into it!

amzid:
Probably what happened here is that the TRD injected the block and did not get any feedback. After a timeout, the transactions are marked as INJECTED and put under done.

The solution here is to correctly parse the output of the payout simulations in order to prevent operation injection in case the simulation (run_operation or preapply) was not successful. In fact, in case of insufficient funds, the run_operation rpc command returns a failure of the operation. We will analyze the reason why this failure is not successfully detected by TRD.

Thank you for reporting the issue.

amzid:
This issue was reformulated in #220 and is solved by #251. Feel free to reopen the issue if needed.

PR #245, opened 2020-08-19
Parse and check result of run_operation

### LINKED ISSUE #244: Check return value of run_operation before injecting
Right now TRD executes the RPC run_operation but does not check the result for possible errors. It should.

utdrmac:
Closes #244

### DIFF
--- a/src/pay/batch_payer.py
+++ b/src/pay/batch_payer.py
@@ -290,6 +290,22 @@ class BatchPayer():
             logger.debug("Error in run_operation, response ->{}<-".format(runops_command_response))
             return PaymentStatus.FAIL, ""
 
+        # Parse result of run_operation and check for potential failures
+        run_ops_parsed = parse_json_response(runops_command_response, verbose=verbose)
+
+        # Check each contents object for failure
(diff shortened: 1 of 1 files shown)


utdrmac:
Effort Time: 3 hours

amzid (approved):
Thanks for the PR. This is a great contribution. Now, the payment fails already at the simulation phase when the balance is too low for example, returns a warning, and does not proceed to the forging and injection steps.

jdsika:
@habanoz 
I think this should be the first candidate for a merge including the transaction hash of the grant payment.

utdrmac:
> Thanks for the PR. This is a great contribution. Now, the payment fails already at the simulation phase when the balance is too low for example, returns a warning, and does not proceed to the forging and injection steps.

Exactly! Should help make things a bit more clearer to users. And there was one guy, @gaia IIRC, who wanted to see the balance beforehand so he could transfer to his payouts address. This should address that in a roundabout way.

jdsika:
@amzid please evaluate if this resolves #116 , #149 and #220

jdsika:
Review and checking: amzid, effort=0,5h

PR #251, opened 2020-08-27
Check the balance of the payout address and warn the user if insufficient

amzid:
IMPORTANT NOTICE: This PR complies with the [guidelines for contribution to the TRD](https://tezos-reward-distributor-organization.github.io/tezos-reward-distributor/contributers.html) and may qualify for being compensated by the TRD grant if approved by the maintainers.

This PR resolves the issues #116, #149 and #220 . The following steps were performed:

* **Bug Analysis**: The balance of the payout address was not checked before performed payments.

* **Implementation**: The balance of the payout address is queried before performing the payment.
    * If the current balance is less than the total payment amount, the payment is immediately stopped and all payment items are marked as failed. The user is notified by a warning message.
    * If the current balance is less than double the total payment amount, the current payment is performed, but the user is notified that the current funds will very likely not be sufficient for the next payment and is asked to transfer funds to the payout address.
    * Add email notifications for both cases. 

* **Performed tests**: the use of different payout addresses with different balances on the test net delivers correct warning messages.

**Work effort**: 2 hrs

### DIFF
--- a/src/api/reward_api.py
+++ b/src/api/reward_api.py
@@ -8,3 +8,7 @@ class RewardApi(ABC):
     @abstractmethod
     def get_rewards_for_cycle_map(self, cycle):
         pass
+
+    @abstractmethod
+    def get_address_current_balance(self, pkh):
+        pass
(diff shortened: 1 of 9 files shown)


amzid:
You are right, thanks for the suggestion, that is a much better solution and will save some code overhead!

jdsika (approved):

Issue #71, opened 2019-04-10
Dynamic PAYMENT_OFFSET


labeoVlad:
**Is your feature request related to a problem? Please describe.**
I am interested in synchronizing my rewards distribution time frame with my next block being baked. In other words I want to be able to pay rewards only during my own block. This way the baker can preserve all the fees generated by his own transactions.  

**Describe the solution you'd like**
To great extent the foundation for this functionality has been already realized.
We already have a so called fixed payment offset 
_-O PAYMENT_OFFSET, --payment_offset PAYMENT_OFFSET
                        Number of blocks to wait after a cycle starts before
                        starting payments. This can be useful because cycle
                        beginnings may be busy._
Now we need to improve it and make it flexible. 
First we need to obtain when next block for the baker is going to happen and wait will that time
Second, we have to execute payments at specified time.   
Would be nice to have it as both : single command line execution and running as a service

**Additional context**
I have analyzed multiple bakers behavior when it comes to make payments. And some of them are already trying to use this strategy. 
Examples are: 
https://tzscan.io/385389
https://tzscan.io/389128
Though, it is still not clear if they do it manually or via script

habanoz:
Researching.

denver-s:
Any update on this issue? I just tried the `payment_offset` and it just didn't work... it should have processed the batch in 5 hours from now but instead it executed right away.

jdsika:
I calssify it as critical as we need to investigate if the payment_offset is either not working or poorly documented.

utdrmac:
In looking at the code, payment_offset is only used when calculating the "sleeping" time to the next payouts period. If you launch TRD with offset 16 and it is currently level_in_cycle 10, it appears to pay out immediately.

This is a bug. payment_offset should be evaluated at the start of each payment loop.

amzid:
> In looking at the code, payment_offset is only used when calculating the "sleeping" time to the next payouts period. If you launch TRD with offset 16 and it is currently level_in_cycle 10, it appears to pay out immediately.
> This is a bug. payment_offset should be evaluated at the start of each payment loop.

PR #259 is a fix for the bug of the missing evaluation of PAYMENT_OFFSET at the start of each cycle. After merging #259, calculations and payments are guaranteed to start only after the configured offset number of blocks.

[...]

PR #252, opened 2020-08-28
Evaluate PAYMENT_OFFSET at the start of each payment loop

utdrmac:
This is a partial fix for #71, which is a feature request for something else but a bug was discovered while investigating.


utdrmac:
Nope. Need to rebase off master branch.

PR #254, opened 2020-08-28
Evaluate PAYMENT_OFFSET at the start of each payment loop

utdrmac:
This is a partial fix for #71, which is a feature request for something else but a bug was discovered while investigating.
Correctly rebased off master this time :)

### DIFF
--- a/src/launch_common.py
+++ b/src/launch_common.py
@@ -83,8 +83,8 @@ def add_argument_release_override(parser):
 def add_argument_payment_offset(parser):
     parser.add_argument("-O", "--payment_offset",
                         help="Number of blocks to wait after a cycle starts before starting payments. "
-                             "This can be useful because cycle beginnings may be bussy.",
-                        default=0, type=int)
+                             "This can be useful because cycle beginnings may be busy.",
+                        default=30, type=int)
 
 
 def add_argument_network(parser):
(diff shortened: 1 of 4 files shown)


jdsika:
PR Template with effort estimation missing + review from @amzid before merge

src/launch_common.py
```diff
@@ -83,8 +83,8 @@ def add_argument_release_override(parser):
 def add_argument_payment_offset(parser):
     parser.add_argument("-O", "--payment_offset",
                         help="Number of blocks to wait after a cycle starts before starting payments. "
-                             "This can be useful because cycle beginnings may be bussy.",
-                        default=0, type=int)
+                             "This can be useful because cycle beginnings may be busy.",
+                        default=30, type=int)
```
jdsika, line 87: I would leave the default value at 0 and only allow uint limited to the maximum number of blocks in a cycle

jdsika (changes requested):

src/launch_common.py
utdrmac: Well, I was hoping we could "help the network" by not making the default be the first block of the cycle each time. I'd really like to make this a random number between 1 and 60 if the user sticks to the default.

utdrmac:
Will re-open PR with template

PR #257, opened 2020-08-28
Check local node bootstrap status

### LINKED ISSUE #249: TRD only works with a fully bootstrapped node before the TRD command gets executed
**Describe the bug**
Unfortunately, TRD does only payout rewards when the Tezos node is fully bootstrapped **before** TRD command gets executed. I discovered the bug accidently and several times by now. 

**To Reproduce**
1. Start Tezos Node 
2. Run Signer
3. Run TRD command and don't wait until the Tezos Node is fully bootstrapped
4. TRD will stuck forever (There is currently no pending reward)

TRD has to be cancled and the command needs to be executed again.

**Expected behavior**
I expect some continous checks like "is the node fully bootstrapped **and** is there a pending reward", if yes execute payout, if no don't execute payout. 

**Environment**
 - Source Code Build
 - Mainnet
 - python3 src/main.py -C 262 -P rpc -M 3 -O 10 -V (usage: payout for payout, shut down node after payout and restart node before payout is due)
- MacOS

**Environment**
I've also expected a similar bug where payouts didn't get executed when the Mac Screen was locked. I think it was the same with screensaver / energy saver.

utdrmac:
Fixes #249 

* **Analysis**: As reported in #249, TRD waits forever if a node is not bootstrapped. TRD should indicate to the user that their node is not bootstrapped and display some progress.
* **Solution**: Monitor bootstrap status of RPC node
* **Implementation**: Fetch /monitor/bootstrap, parse timestamp, check within 2 minutes of local time, otherwise sleep 8 blocks and repeat.
* **Performed tests**: See screenshot of tested setup
* **Documentation**: No doc changes necessary

* **Check list**:

[NA] I extended the Travis CI test units with the corresponding tests for this new feature (if needed).
[NA] I extended the Sphinx documentation (if needed).
[NA] I extended the help section (if needed).
[NA] I changed the README file (if needed).
[NA] I created a new issue if there is further work left to be done (if needed).

**Work effort**: 12hrs


jdsika:
I think it is the right approach! `node_is_bootstrapped(self)`

And with the restart loop and the sleep timer it should work!

jdsika:
@utdrmac CI fail, synthax

src/pay/payment_producer.py
```diff
@@ -22,6 +22,8 @@
 logger = main_logger
 
 MUTEZ = 1e+6
+BOOTSTRAP_SLEEP = 32
```
jdsika, line 25: why 32?

dansan566:
```
Exception in thread producer:
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/Users/danielliebert/tezos-reward-distributor/src/pay/payment_producer.py", line 124, in run
    time.sleep(5)
NameError: name 'time' is not defined
```

jdsika:
I am also getting this error: 

```
  File "src/main.py", line 220, in <module>
    main(args)
  File "src/main.py", line 149, in main
    recent = get_latest_report_file(payments_root)
  File "src/main.py", line 206, in get_latest_report_file
    key=lambda x: int(x))
  File "src/main.py", line 206, in <lambda>
    key=lambda x: int(x))
ValueError: invalid literal for int() with base 10: '_326_'
```

src/pay/payment_producer.py
utdrmac: semi random pick. Assuming a new node would need several hours, potentially, to bootstrap, I did not want the loop-check happening frequently. 32 blocks is just over 30m; just trying to keep it a power of 2 for no real reason. I'm open to something less if you think we should check more frequently.

utdrmac:
>   File "/Users/dl/tezos-reward-distributor/src/pay/payment_producer.py", line 124, in run
>     time.sleep(5)
> NameError: name 'time' is not defined
> ```

I'm curious why the TravisCI check did not encounter this.

utdrmac:
@jdsika You had something different happen:
`Aug 30 11:49:46 baker4 python3[24738]: NameError: name 'node_is_bootstrapped' is not defined`

Which caused your TRD to exit abruptly, probably causing that exception

src/pay/payment_producer.py
jdsika: help me out here: I thought we are coping with the fact that the "bootsrapped" command in tezos-client returns even if it is not bootstrapped yet. For this reason you will call the command again and again through a loop and compare the time message in the return value with the local time of the machine. This is the logic that I would expect from the tezos-client btw...
As long as the bootstrapping is not finished it would be nice to notify the user of the progress that is being made. Therefore the bootstrap loop could call itself every after a new block has been found and a new time value can be compared. I would use "2" in this case?

dansan566:
2020-08-31 16:07:54,711 - producer  - Error at payment producer loop
  File "/Users/dl/tezos-reward-distributor/src/pay/payment_producer.py", line 129, in run
    if not node_is_bootstrapped(self.node_url):
NameError: name 'node_is_bootstrapped' is not defined
2020-08-31 16:07:59,720 - producer  - Error at payment producer loop
  File "/Users/dl/tezos-reward-distributor/src/pay/payment_producer.py", line 129, in run
    if not node_is_bootstrapped(self.node_url):
NameError: name 'node_is_bootstrapped' is not defined
2020-08-31 16:08:04,724 - producer  - Error at payment producer loop
  File "/Users/dl/tezos-reward-distributor/src/pay/payment_producer.py", line 129, in run
    if not node_is_bootstrapped(self.node_url):
NameError: name 'node_is_bootstrapped' is not defined

utdrmac:
@dansan566 Sorry. Please check again. It bugs me that Python cannot detect these kinds of issues via unittests nor via normal syntax checking.

src/pay/payment_producer.py
utdrmac: Yes, /bootstraped always returns. Seems to be a stream of data even when using `curl` which I did not realize. I thought that if you used `curl` (aka python requests) that it would fetch 1 entry and exit. So I need to modify the code to handle this.

Yes, this is more/less the logic of the tezos-client but I'm trying to remove usage/dependency of the tezos-client.

I can certainly modify the code to print out more frequently, loop more often.

utdrmac:
Ok @dansan566 @jdsika Finally got a chance to test this myself. I started a new carth docker instance and while it was bootstrapping I was able to test this out. See screenshot of log output.
Screen Shot 2020-08-31 at 11 41 33 AM

> help me out here: I thought we are coping with the fact that the "bootsrapped" command in tezos-client returns even if it 
> is not bootstrapped yet.

Turns out, even with `curl` the /monitor/bootstrapped is a stream of data that does not end until node is bootstrapped. So I have it set to just grab/parse a few lines of output to check, then it sleep-waits and repeats.

> This is the logic that I would expect from the tezos-client btw...

Yes, but I'm trying to remove all usage of tezos-client. If we want better support for Kiln/Docker/etc, we need to stop calling the tezos-client binary directly and do everything through http/rpc.

> As long as the bootstrapping is not finished it would be nice to notify the user of the progress that is being made.

Done. Changed it to 8 blocks/minutes of waiting between checks so that messages are a bit more frequent.

jdsika:
We need to also use the PR template here and your work effort

dansan566:
```
AttributeError: 'TzKTBlockApiImpl' object has no attribute 'get_bootstrapped'
2020-09-01 23:27:40,600 - producer  - Error at payment producer loop
  File "/Users/dl/tezos-reward-distributor/src/pay/payment_producer.py", line 130, in run
    if not self.node_is_bootstrapped():
  File "/Users/dl/tezos-reward-distributor/src/pay/payment_producer.py", line 307, in node_is_bootstrapped
    boot_time = self.block_api.get_bootstrapped()
```

utdrmac:
@dansan566 Tested myself using `-P rpc` and `-P tzkt`. Should be working well now.

dansan566:
Worked! Love it - great work, thanks!

src/pay/payment_producer.py
```diff
@@ -122,6 +125,15 @@ def run(self):
             time.sleep(5)
 
             try:
+
+                # Check if local node is bootstrapped; sleep if needed; restart loop
+                if not self.node_is_bootstrapped():
+                    logger.info("Local node, {}, is not in sync with the Tezos network. Will sleep for {} blocks and check again."
+                                .format(self.node_url, BOOTSTRAP_SLEEP))
+                    self.wait_until_next_cycle(BOOTSTRAP_SLEEP)
```
amzid, line 133: Should this not be `self.wait_for_blocks(BOOTSTRAP_SLEEP)` instead?

utdrmac: Yes it should

amzid (approved):

jdsika:
@utdrmac can you retrigger the CI? I don't know why it is not passing the results.
@amzid how long did the review take?

I will merge when the status in green

amzid:
> @amzid how long did the review take?

0.5 h

> @utdrmac can you retrigger the CI? I don't know why it is not passing the results.

I think it is not activated for the branch (fork), but it passed for the PR.

PR #259, opened 2020-08-31
Evaluate PAYMENT_OFFSET at start of each cycle

utdrmac:
This PR resolves the issue #71 . The following steps were performed:

* **Analysis**: Confirmed bug report by user in issue
* **Solution**: Change code to evaluate current level-in-cycle (0-4095) vs OFFSET to ensure that payouts do not happen before OFFSET 
* **Implementation**: ^^
* **Performed tests**: Modified code and self-tested using `-O 1033` on current cycle, which was at level 1028. Outputted correct information.
* **Documentation**: No changes to docs needed

* **Check list**:

[NA] I extended the Travis CI test units with the corresponding tests for this new feature (if needed).
[NA] I extended the Sphinx documentation (if needed).
[NA] I extended the help section (if needed).
[NA] I changed the README file (if needed).
[NA] I created a new issue if there is further work left to be done (if needed).

**Work effort**: 8hrs


src/launch_common.py
```diff
@@ -84,8 +84,8 @@ def add_argument_release_override(parser):
 def add_argument_payment_offset(parser):
     parser.add_argument("-O", "--payment_offset",
                         help="Number of blocks to wait after a cycle starts before starting payments. "
-                             "This can be useful because cycle beginnings may be bussy.",
-                        default=0, type=int)
+                             "This can be useful because cycle beginnings may be busy.",
+                        default=30, type=int)
```
jdsika, line 88: see comment from last time. The user will not expect an offset by default and we have to use range [-30, XXX] and I would like to use uint from 0 to cycle end for safety

src/launch_common.py
```diff
@@ -83,11 +82,18 @@ def add_argument_release_override(parser):
                         default=0, type=int, action=ReleaseOverrideAction)
 
 
+class PaymentOffsetAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        if not (values >= 0 and values < 4096):
```
jdsika, line 87: 4096 should be a global constant as it is the number of blocks per cycle. No magic numbers

jdsika:
@amzid the conflict resides from the bug in the logger, { :s } formates into a string?

amzid:
@jdsika yes this is what I think, let's wait until these whole new changes get tested more intensively, the bug is not related to this PR though!

amzid:
> @amzid the conflict resides from the bug in the logger, { :s } formates into a string?

@jdsika  Yes!



> * **Solution**: Change code to evaluate current level-in-cycle (0-4095) vs OFFSET to ensure that payouts do not happen before OFFSET

@utdrmac thanks for the PR. This PR guarantees that the calculations and payouts do not start before the offset. Do you think there is a solution to make the payment happen exactely at the requested offset as described in #71?

utdrmac:
@amzid Yes, we could add something so that TRD waits for a specific block/level, but that would be lots of constant checking /head (every 10s or so) to match the level. TRD would have to process everything (getting balances, etc) before the requested level. When the request level arrives, then we preapply/forge/sign/inject.

jdsika:
We currently call it offset in blocks which is implying we are able to payout after waiting a specific amount of blocks which cannot be guranteed as I undestand?
In this case I woul either implement it as amzid proposes or change into a time amount?

utdrmac:
@jdsika The original way this parameter was used was to determine how long to sleep until next payouts. Say you paid C123 at level 44 with an offset of 16. TRD would do sleep(((4095-44)+16)*60s) So it wasn't ever a true "block offset" to begin with, especially considering that some blocks might get prio1 bake, thus taking longer than 60s.

With the changes in this PR, that logic still exists for calculating a sleeping period, and additional it is more specific to looking at the level within the cycle compared to the offset parameter.

Being able to send the trx at a specific chain level would require more code changes. All of the "bulk calculations" like fetching balances, rewards, etc would need to be done several levels in advance. I have 300+ delegators and it takes much longer than 60s to fetch all their balances so it all could not be done at the level I want.

amzid (approved):
The changes were successfully tested.
Considering the related overhead of implementing the ability of injecting the transaction at a specific level, I think we should have more discussions about the need/prioritization of this feature.

utdrmac:
> Considering the related overhead of implementing the ability of injecting the transaction at a specific level, I think we should have more discussions about the need/prioritization of this feature.

Agreed. This PR fixes the initial issue. We can do another feature request for specific level payouts.

Issue #266, opened 2020-09-04
NameError: name 'loggin' is not defined


songproducer:
```
Traceback (most recent call last):
  File "src/main.py", line 223, in <module>
    main(args)
  File "src/main.py", line 40, in main
    loggin.info("Anonymous statistics {} be collected. See docs/statistics.rst for more information."
NameError: name 'loggin' is not defined

```

jdsika:
thx @adamkittelson for the fix

PR #267, opened 2020-09-06
fix typo, addresses issue #266

adamkittelson:
fixed a typo preventing the service from starting, addresses issue #266


utdrmac:
Why didn't TravisCI catch this?

Issue #278, opened 2020-10-10
Exception: Lock file present. Another process is running...


k-vyn:
I keep getting when I run the script: 
```
2020-10-10 01:28:33,474 - MainThread - --------------------------------------------
Traceback (most recent call last):
  File "src/main.py", line 225, in <module>
    main(args)
  File "src/main.py", line 146, in main
    life_cycle.start(not dry_run)
  File "/Users/-/tezos-reward-distributor/src/util/process_life_cycle.py", line 20, in start
    self.lock_file.lock()
  File "/Users/-/tezos-reward-distributor/src/util/lock_file.py", line 10, in lock
    self.tryLock()
  File "/Users/-tezos-reward-distributor/src/util/lock_file.py", line 20, in tryLock
    raise Exception("Lock file present. Another process is running...")
Exception: Lock file present. Another process is running...
```
Not sure how this happened, but I have to imagine the last script didn't end neatly. How do I go about reseting this? Can't seem to find a command.

Thanks!

dansan566:
@k-vyn We had the same issue once. Quick fix: delete the lock file in the trd folder

jdsika:
@amzid should we add a command to delete the lock file with a warning to only do it when the application crashed?

k-vyn:
I tried deleting the lock file, but I then get this 

```
  File "src/main.py", line 225, in <module>
    main(args)
  File "src/main.py", line 152, in main
    recent = get_latest_report_file(payments_root)
  File "src/main.py", line 210, in get_latest_report_file
    files = sorted([os.path.splitext(x)[0] for x in os.listdir(get_successful_payments_dir(payments_root))],
  File "src/main.py", line 211, in <lambda>
    key=lambda x: int(x))
ValueError: invalid literal for int() with base 10: '.DS_Store  File "src/main.py", line 225, in <module>
    main(args)
  File "src/main.py", line 152, in main
    recent = get_latest_report_file(payments_root)
  File "src/main.py", line 210, in get_latest_report_file
    files = sorted([os.path.splitext(x)[0] for x in os.listdir(get_successful_payments_dir(payments_root))],
  File "src/main.py", line 211, in <lambda>
    key=lambda x: int(x))
ValueError: invalid literal for int() with base 10: '.DS_Store
```

k-vyn:
I also tried deleting TRD and re-cloning it, but I get the same error. Should I re-create my pymnt folder?

[...]

PR #285, opened 2020-10-24
Allow deleting lock file if no other process is running

amzid:
This PR resolves the issue #278. It allows deleting the lock file if no other process is running. This should be forced by the user by entering 'y'.

**Work effort**: 0.5h


src/tzstats/tzstats_reward_provider_helper.py
```diff
@@ -176,7 +176,7 @@ def split(input, n):
     def update_current_balances(self, reward_logs):
         """External helper for fetching current balance of addresses"""
         for rl in reward_logs:
-            rl.current_balance = self.__fetch_current_balance(rl.address)
+            rl.current_balance = self.__fetch_current_balance([rl.address])[rl.address]
```
jdsika, line 179: What is this change here?

jdsika (approved):

PR #293, opened 2020-11-07
Update version to match release version

utdrmac:
Simple fix.
Effort: 0


utdrmac:
@jdsika Can you merge this and then update the release tag?

jdsika (approved):
good catch. We have to check for this before the release

Issue #209, opened 2020-03-05
Task: Remove direct usage of tezos-client


utdrmac:
**Describe the task**
There are many instances in TRD that make shell/command calls to execute the `tezos-client` and parse the result. Most, if not all, of these are unnecessary and the generic HTTP API should be used instead with the python requests library. Most `-client` calls become HTTP requests anyways, so this isn't that big of a move.

**Reasoning**
The `tezos-client` has massive overhead for each execution. By removing these shell calls, we improve TRD responsiveness and remove a dependency.

Because TRD encourages the use of tezos-signer, TRD can make all necessary signing calls directly to tezos-signer. TRD only needs to know the PKH, it does not need to know any wallet aliases.

Issue #247, opened 2020-08-22
Investigate why CLI access to tezos-client is required (Kiln Support)


MHULT:
Hey guys,

Has someone got TRD working together with Kiln?

If so... how?

Best regards,
M

utdrmac:
Unfortunately, due to the way Kiln hides all of the tezos binaries, it is currently not possible to make them work together. There is an open kiln issue on this: https://gitlab.com/obsidian.systems/kiln/issues/16
There are some crazy-hacky workarounds in that issue if you want to take a stab at it.

Will leave this issue open as FR into researching what TRD really needs and whether or not accessing the binaries is critical.

jdsika:
@utdrmac is my assumtion correct, that the RPC is left open on the standard port and localhost for kiln so that the removal of the tezos binary dependency would solve this issue?

utdrmac:
@jdsika That is my understanding yes.

jdsika:
@amzid I am assigning you, as it is a synergy for planned work automatically solved with "removing the tezos binary dependency". Label has been changed to "blocked" only.

PR #294, opened 2020-11-09
[REVIEW_NEEDED] Remove dependency from tezos client

amzid:
This PR resolves the issue #209 . The following steps were performed:

* **Analysis**: The Tezos client is used to do rpc get and post requests, wait for confirmations, sign transactions and get known addresses.

* **Implementation**: The dependency for rpc get and post requests, waiting for confirmations was removed. 
UPDATE: The dependency from the tezos client for the signing process and known addresses was performed.

* **Performed tests**: Test on the carthagenet and delphinet. The unit tests were adjusted.

* **Documentation**: The help section was adjusted. 

**Work effort**: 28 hrs.


utdrmac:
This looks good so far. The biggest issue, IMO, will be all of the "wallet" functions.

* At startup, TRD runs "list known addresses" and attempts to resolve address aliases to actual addresses. This functionality is not available via RPC because these are tezos-client wallet functions.
* There is no RPC for "sign bytes", because this is wallet function.
* I think deeper integration with tezos-signer will be required if we truly want to remove all usage of tezos-client. You can query tezos-signer and request a list of keys that it knows, and you can POST to signer for signing bytes.

habanoz:
@amzid 

Why do we need to remove tezos-client dependency?

habanoz:
@amzid @utdrmac @jdsika 

Well. I wonder why we need to remove client dependency and I saw the linked issues. 

I believe this task comes with a lot of effort but brings very little back. 

The responsiveness of the TRD application is not affected by a degree that affects user experience. After all, TRD does not need to be lighting fast. It runs once in a cycle (3 days). 

Tezos client is available almost everywhere where a tezos blockchain is present. I admit it is not a very handsome way to make shell calls to the client, however, implementing a tezos client in trd should not be a replacement.

For the KILN case, let's assume #247 is resolved with this approach. How many users will make use of this solution? How common is the use of KILN? 

If there are any benefits I cannot see please help me see.

jdsika:
I will put it this way:
If Ahmed pulls it off in a few days I want the dependency gone. If he tells me after two days it is too hard. I reconsider the whole thing. We have quite a few issues with people using wrong binaries etc.. I am always "pro removing weight"

amzid:
> * At startup, TRD runs "list known addresses" and attempts to resolve address aliases to actual addresses. This functionality is not available via RPC because these are tezos-client wallet functions.

I will make the possibility of using alias only possible if a tezos-client is used.
> * I think deeper integration with tezos-signer will be required if we truly want to remove all usage of tezos-client. You can query tezos-signer and request a list of keys that it knows, and you can POST to signer for signing bytes.

I could not find any resources or documentation on how to interact with the tezos signer using http requests, can you find any?

amzid:
I would like to review and merge the current changes which remove the dependency from the client for get and post rpc requests. I opened an [issue](https://gitlab.com/tezos/tezos/-/issues/1035) about an easier way to sign an operation without using the client. Once we tackle this, we could make TRD completely independent from the client.

utdrmac:
@amzid All you have to do is make a POST to `/keys/<pkh>` with the bytes you want to sign and signer signs it and returns a JSON string with 1 member "signature". It's quite easy. You can make GET request to the same URL `/keys/<pkh>` to fetch their public key and verify that the signer can sign for their address.

jdsika:
Ok, let's split the work. I want this one not to be open too much longer. @amzid has the todo for the removal of the tezos-client for signing

jdsika:
@amzid please make sure all dependencies and their recommended way of installation (OPAM) are mentioned correctly in the readme

This prefix is needed btw: **OPAMSOLVERTIMEOUT=600** opam depext tezos

amzid:
> @amzid please make sure all dependencies and their recommended way of installation (OPAM) are mentioned correctly in the readme
> 
> This prefix is needed btw: **OPAMSOLVERTIMEOUT=600** opam depext tezos

there will be some changes due to #357 and #361. Maybe we can postpone the update of the Readme until the exact steps, configs and dependencies are fully defined and more structured.

habanoz (commented):
- Please do not use verboseOverride. It is better to remove all verbose logging related control structures. 
- wllt_clnt_mngr needs to be renamed to clnt_mngr as it is done almost everywhere
- Unable to find a unit test for ClientManager functionalities. If there is please ignore, if there is none please add one.
- Please move consecutive logger lines and exception statements expanding to multiples to separate methods for the sake of readability
- Please make sure all HTTP calls are persistent. HTTP calls should either be retried a predefined number of times e.g. 5 even better is using a configuration specified amount. If all http calls are moved to a single class, implementing this behavior will be simple.
- Client executable parameter is removed. I can see signing related code sections but I cannot see any parts about the block injection. Have you completely dropped tezos client? If so how do you implement block validation and injection?
- Nice work!
- Any questions? I will be available in telegram.
- I urge you to plan a task to review all code base and refactor code to increase readability.

amzid:
> * Please do not use verboseOverride. It is better to remove all verbose logging related control structures.
> * wllt_clnt_mngr needs to be renamed to clnt_mngr as it is done almost everywhere

done in 60221b3

> * Please move consecutive logger lines and exception statements expanding to multiples to separate methods for the sake of readability
> * Please make sure all HTTP calls are persistent. HTTP calls should either be retried a predefined number of times e.g. 5 even better is using a configuration specified amount. If all http calls are moved to a single class, implementing this behavior will be simple.

done in b51b333

> * Client executable parameter is removed. I can see signing related code sections but I cannot see any parts about the block injection. Have you completely dropped tezos client? If so how do you implement block validation and injection?

Both are done in batch payers with POST http requests. For the injection, the signed operation bytes are given as parameters to the POST request in json format. For the validation, the five blocks after the block injection are queried and the op hash is fetched in the op hash lists, if it is found, the batch payer returns PAID status, otherwise INJECTED.

> * Nice work!
> * Any questions? I will be available in telegram.

Thank you for the input!

> * I urge you to plan a task to review all code base and refactor code to increase readability.
> * Unable to find a unit test for ClientManager functionalities. If there is please ignore, if there is none please add one.

The code refactoring and missing tests will be part of v9.0! :)

jdsika (approved):
Approve after habanoz review

PR #300, opened 2020-11-12
Documentation Updates

utdrmac:
**Work effort**: 1hr


docs/conf.py
```diff
@@ -49,17 +49,17 @@
...
+copyright = u'2020, habanoz'
+author = u'TRD Organization'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = u'1.0'
+version = u'7.0'
```
jdsika, line 60: I think we need to merge 8.0 as this is the next release

docs/conf.py
```diff
@@ -49,17 +49,17 @@
 
 # General information about the project.
 project = u'Tezos Reward Distributor'
-copyright = u'2019, habanoz'
```
jdsika, line 52: the date is normally stating since when a copyright is claimed, changing it would mean that the copyright from 2020 onwards and not before

jdsika (changes requested):

docs/conf.py
utdrmac: oh, ok. No worries.

docs/conf.py
utdrmac: How about a 7.1 for all the minor fixes we've done since 7.0?

jdsika: I will introduce a proper version numbering. I propose a major version change for every protocol change. Minor changes for improvements within a protocol. We aim for a cadence to release major with every protocol activation and a X.5 in the middle of the protocol phase. As NL announced a 3 month rhythm for protocols we would aim for a release every 6weeks.

jdsika:
you need to start new branches fresh from master to not have this commit message mess :D

jdsika (approved):

PR #301, opened 2020-11-12
If supporters_set exists but not set to anything, create empty set

utdrmac:
**Work effort**: 0


utdrmac:
@dansan566 for BitCat's issue

jdsika (approved):

PR #306, opened 2020-11-14
Update twitter.py

utdrmac:
- Corrected logic to determine if parameter is list or not
- fixed unitesting for delphinet

Work Effort: 1hr


jdsika (approved):

utdrmac:
The build is failing because the RPC we are using for test networks switched over to the delphi test net which doesn't have the original test baker.

PR #319, opened 2020-11-20
More unit tests, introduce Pytest

utdrmac:
* **Analysis**: Several aspects of TRD do not have unittests for discovering issues prior to releases
* **Solution**: Nearly every function in TRD should have a unit test of some kind to make sure nothing breaks
* **Check list**:
- [X] I extended the Travis CI test units with the corresponding tests for this new feature (if needed).

**Work effort**: 16hrs

I'd like to move all unit tests into a common folder at the root project `tests/` Any objections?


jdsika:
@utdrmac I am in favor of moving all tests to a subfolder `/tests/`

Should we use [Pytest](https://docs.pytest.org/en/stable/) ?

src/NetworkConfiguration.py
```diff
@@ -8,7 +8,7 @@
     'MAINNET': {'NAME': 'MAINNET', 'NB_FREEZE_CYCLE': 5, 'BLOCK_TIME_IN_SEC': 60, 'BLOCKS_PER_CYCLE': 4096,
                 'BLOCKS_PER_ROLL_SNAPSHOT': 256},
     'ALPHANET': {'NAME': 'ALPHANET', 'NB_FREEZE_CYCLE': 3, 'BLOCK_TIME_IN_SEC': 30, 'BLOCKS_PER_CYCLE': 2048,
```
jdsika, line 10: Maybe we should rename those accrding to how tezos client/protocol devs are handling the naming? We release with every new protocol version and make it a "pre release todo" to rename those according to the new names?

src/NetworkConfiguration.py
```diff
@@ -8,7 +8,7 @@
...
-                 'BLOCKS_PER_ROLL_SNAPSHOT': 256},
+                 'BLOCKS_PER_ROLL_SNAPSHOT': 128},
```
jdsika, line 11: This change means we have a bug in the current release?

jdsika:
Prioritizing this for v8.1 release

src/NetworkConfiguration.py
utdrmac: I'm ok with that. It does make things confusing to say `-N ALPHANET` when you want to test against delphinet. Do we need to keep zeronet? Does anyone use that?

src/NetworkConfiguration.py
utdrmac: Yes, unfortunately. https://delphinet-tezos.giganode.io/chains/main/blocks/head/context/constants

utdrmac:
> Should we use [Pytest](https://docs.pytest.org/en/stable/) ?

Well, all of the current tests are written using unittest and Mock. So we would have to rewrite all our tests to use pytest.

src/NetworkConfiguration.py
amzid: I think zeronet is not running right now (http://tezos.gitlab.io/introduction/test_networks.html). Let's keep these always updated only with the new and running test networks!

amzid:
I am not sure if the current tzkt tests are still up-to-date? At least from the test file names, some old cycles are being marked as expected so it might be better to adjust the names and check if they need to be updated?

Maybe we can think about more tests for the future which fetch data from all providers and compare the fetched data based on some similarity metrics. The test fails for some exceeded discrepancy threshhold.

utdrmac:
@amzid One thing that I like about the tzkt tests is that it actually launches/runs main.py and "runs" TRD with a custom config. This is something we really need. I'm simulating the entire payouts process in one of the tests I added in this PR that does everything payments_consumer/producer do. But I'd really like to expand that and run "main" for each data provider.

src/NetworkConfiguration.py
jdsika: Please rename accoring to the conventions used in the Tezos repository. We maintain the networks: carthagenet, delphinet, etc and depricate them accorind to the shut down plans. e.g. carthage will be shut down 12.12.2020.

jdsika:
> "Pytest is designed to be a simpler, easier-to-code alternative to unittest. Pytest provides essentially the same features as nose, but is better supported and more widely used. Unlike doctest and unittest, pytest is not part of the Python standard library"
[Why is Pytest more popular](https://blog.j-labs.pl/2019/02/Pytest-why-its-more-popular-than-unittest)

According to @amzid unittests are compatible with pytest. We could vote to keep the current tests but do all new ones with pytest. If we rewrite an old one we migrate to pytest. 
Please state if you would be in favour of this proposal. @utdrmac @habanoz

utdrmac:
@jdsika This PR is getting to big for my comfort so I'm going to stop here. I moved all unit tests into tests/ and updated the travis to run against them in there. I also cleaned up some unused code, started expanding the logging, and added a few more try/except that i discovered with my new unit tests.

* The tzkt tests still work, as far as i can tell.
* test_phases.py is a new one i created, that simulates the code path within payment-producer/consumer. This test is mainly to make sure RPC calls continue to work and that all the math behind rewards remains consistent. There's also some mapping and merging tests too.
* test_calculatePhaseMerge simply tests if merging multiple payments to the same address works correctly
* test_rpc_api is true execution of src/main.py, modeled after tzkt's testing suite. This test requires network access and actually fetches data from delphinet. There's currently no comparisons to check that the data is correct, but this is a start to making sure config is processed, data is fetched, etc.

jdsika:
OK, So let us wrap this up. The next step would be @amzid to finalize and review the PR with care as there are a lot of changes. Please take your time until Saturday. I will merge then and release v8.1

amzid:
The first unit test is failing:
2020-11-28 16:39:54,311 - main.payment_producer - Producer "producer" started
Exception in thread producer:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/it-test/tezos-reward-distributor/src/pay/payment_producer.py", line 139, in run
    get_verbose_log_helper().reset(pymnt_cycle)
AttributeError: 'NoneType' object has no attribute 'reset'

amzid (approved):

jdsika (approved):

habanoz:
@utdrmac @amzid 

VerboseLoggerHelper needed to be initialized. For the sake of unit tests, I added a dummy VerboseLoggerHelper instance.

PR #320, opened 2020-11-21
Swap pycodestyle for flake8 to include code linting

utdrmac:
The flake8 utility is a combination of _pycodestyle_ and _pyflakes_.

> pyflakes is a verification tool(linter) which checks for Python files for errors. Pyflakes doesn’t verify the style at all but it verifies only logistic errors like the syntax tree of each file individually.

* **Check list**:

- [X] I extended the Travis CI test units with the corresponding tests for this new feature (if needed).

**Work effort**: 2hrs


.flake8
```diff
@@ -1,3 +1,8 @@
 [pycodestyle]
```
amzid, line 1: Is the pycodestyle configuration still needed even though the flake8 style is used instead now?

utdrmac: No it is not. Left it in accidentially.

amzid (approved):

Issue #352, opened 2020-12-25
gas exhausted operation


Primate411:
**Describe the bug**
Paying cycle 309 resulted in an error. This is the first payment for delegator [](https://tzstats.com/KT1UcPh3S1K1GAFqUt212H9qjFVdxEnca1qk)

Actual error:
Error while validating operation - Status: failed, Message: proto.007-PsDELPH1.gas_exhausted.operation

**To Reproduce**
Running the following command which is what I usually run to pay frozen cycles.
[app_verbose_309_20201223_204000.log](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/files/5740967/app_verbose_309_20201223_204000.log)

python3 src/main.py -E ~/tezos -M 3 -C 309 -R -5 -P tzstats -V on

Using default fee.ini or using fee.ini from last rev in January.

**Expected behavior**
The fee automatically calculates properly to be included in baking but not be too high to be wasteful.

**Screenshots**
N/A

**Environment**
 - Latest source coce
 - Attached configuration (note: API info removed, attached as text
[tz1S5WxdZR5f9NzsPXhr7L9L1vrEb5spZFur.txt](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/files/5740971/tz1S5WxdZR5f9NzsPXhr7L9L1vrEb5spZFur.txt))
 - Mainnet
 - python3 src/main.py -E ~/tezos -M 3 -C 309 -R -5 -P tzstats -V on
 - Ubuntu 20.10

**Additional context**
N/A

Thank you

Primate411:
I got around this problem for now by redirecting the KT1's rewards to its manager tz1 address.  So far 2 cycles this way no problem.

jdsika:
We will make this redirect most likely the default behavior and bundle those transactions belonging to the same manager tz

jdsika:
fixed with #353

PR #353, opened 2020-12-29
Add support for delegated KT accounts with active default entrypoint

amzid:
This PR implements support for delegated KT accounts with active default entrypoints. The TRD calculates the simulated consumed gas and sets the limit for gas and the baker fee on the basis of the calculated amount.

This PR is a partial fix for #352: It covers the delegated contracts which have a default entrypoint. If the delegated account does not have a default entrypoint, the user should forward the payment to another supported payment address using the rules_map.

UPDATE: add storage cost computation + bound total fee amount.

* **Performed tests**: Test payouts to vaults on Delphinet.

**Work effort**: 7 hrs

### DIFF
--- a/src/pay/batch_payer.py
+++ b/src/pay/batch_payer.py
@@ -35,6 +35,11 @@ FEE_INI = 'fee.ini'
 RA_BURN_FEE = 257000  # 0.257 XTZ
 RA_STORAGE = 300
 
+# This fee limit is set to allow payouts to ovens
+# Other KT accounts with higher fee requirements will be skipped
+# TODO: define set of known contract formats and make this fee for unknown contracts configurable
+FEE_LIMIT_CONTRACTS = 12444
+
 # For simulation
 HARD_GAS_LIMIT_PER_OPERATION = 1040000
 HARD_STORAGE_LIMIT_PER_OPERATION = 60000
@@ -263,8 +268,9 @@ class BatchPayer():
                 self.wait_random()
 
         for payment_item in payment_items:
-            payment_item.paid = status
-            payment_item.hash = operation_hash
+            if payment_item.paid == PaymentStatus.UNDEFINED:
+                payment_item.paid = status
+                payment_item.hash = operation_hash
 
         return payment_items, attempt_count
 
@@ -297,7 +303,7 @@ class BatchPayer():
             logger.debug("Error in run_operation, request ->{}<-".format(runops_command_str))
             logger.debug("---")
             logger.debug("Error in run_operation, response ->{}<-".format(runops_command_response))
-            return PaymentStatus.FAIL, ""
+            return PaymentStatus.FAIL, []
         run_ops_parsed = parse_json_response(runops_command_response)
         op = run_ops_parsed["contents"][0]
         status = op["metadata"]["operation_result"]["status"]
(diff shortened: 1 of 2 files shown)


amzid:
A review from the side of @maht0rz suggests the following:
1) the implementation should cover potential extra storage requirements
2) We should assert that the resulting fee (with extra gas and storage costs) should remain below the reward amount.
3) Reward amounts for multiple cycles can be aggregated to have a greater value than the required fee. This is even more important for KT1 accounts.
4) With the current gas estimation for vaults and the current block gas hard limit for delphi, the batch size cannot exceed 100. The batch size would be should be computed dynamically,

The first two suggesions will be implemented in this PR. For the two remaining points (3 & 4), extra issues will be opened.

utdrmac (approved):

dexterslabor:
Hi @amzid, at Stove Labs we are developing an SDK for interacting with vaults/ovens of the wXTZ project. 
It has a method implemented to check for contract code integrity. Maybe this is of any use to identify correctly oven contracts.

Fetching
https://github.com/stove-labs/wrapped-xtz-sdk/blob/dev/src/WXTZBase.ts#L34-L39
Checksum
https://github.com/stove-labs/wrapped-xtz-sdk/blob/dev/src/deployments.json#L24
Checking
https://github.com/stove-labs/wrapped-xtz-sdk/blob/dev/src/byteHelpers.ts#L17-L20

jdsika:
@amzid is this ready to be reviewed now? Are all comments considered?

amzid:
#357 is opened for remaining work

src/pay/batch_payer.py
```diff
@@ -292,13 +347,26 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
...
 
+            # TRD extension for non scriptless contract accounts
+            gas_limit, tx_fee = self.gas_limit, self.default_fee
+            if payment_item.paymentaddress.startswith('KT'):
+                simulation_results = self.simulate_single_operation(payment_item, pymnt_amnt, branch, chain_id)
+                gas_limit, tx_fee, storage = simulation_results
+                burn_fee = COST_PER_BYTE * storage
+                total_fee = tx_fee + burn_fee
+                # Bound the total (baker and burn) fee by the reward amount in case of KT1 accounts
+                if total_fee > pymnt_amnt:
```
jdsika, line 358: where is the connection to the limits set in the fee.ini?

amzid: The initial fee value is taken from fee.ini. The extra fee needed for gas + extra burn fee needed for storage are simulated and adjusted according to the simulation output. The values for tz1 accounts are taken directly from fee.ini.

jdsika: the boundary if(fee > reward) is not sufficient. we need maximum limits services are willing to pay. Idea: create the fee.ini at first startup. Add the name of the current protocol. check it at startup and update with new fees if protocol changes. Store the simulated values for a tz account and add max_fees to the ini as a fail criteria. The Kt accounts with internal calls can be used to trick services into paying large amounts of fees.

jdsika:
Initial state: fee.ini is not part of the repository
1) Check if fee.ini exists
-> no: create fee.ini, fetch hard limits from protocol, fetch protocol verison, simulate tz account costs and write to fee.ini
-> yes: check if fee.ini protocol version is the same as the currently used one. -> if no: delete fee.ini and initialize again

jdsika:
@amzid I think you still need to change the parts we discused before @utdrmac can do a final approval?

amzid:
> @amzid I think you still need to change the parts we discused before @utdrmac can do a final approval?

I would like to close this PR here. This includes the capability of paying out to KT1 contracts with default entrypoints with a fee limit adjusted on basis of the oven code (which should be the most costly KT contract known to TRD up to now). The identification of KT accounts and adjustable fees according to protocol/contract is opened in a separate issue described in #357.

jdsika (approved):
LGTM

PR #358, opened 2021-01-24
Dexter fee ratio

amzid:
This PR fixes the dexter implementation in case rewards for Dexter contract are zero (or if configured contract not delegated to baker).
Furthermore, this PR adds the possibility to configure a global fee ratio in special maps for all accounts inferred from the Dexter account (LPs). These LPs will share the fee ratio of the Dexter contract configured in specials map.

* **Performed tests**: Test future rewards for a sample baker with one/multiple LPs and compare results with and without Dexter configuration. 

**Work effort**: 2.5 hrs.

### DIFF
--- a/src/Dexter/dexter_utils.py
+++ b/src/Dexter/dexter_utils.py
@@ -33,17 +33,6 @@ def process_original_delegators_map(delegator_map, contract_id, snapshot_block, 
             delegator_map[delegator]['current_balance'] = dexter_liquidity_provider_map[delegator]['current_balance']
         delegator_map[delegator]['originaladdress'] = contract_id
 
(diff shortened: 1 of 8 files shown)


src/Dexter/dexter_utils.py
```diff
@@ -31,6 +31,7 @@ def process_original_delegators_map(delegator_map, contract_id, snapshot_block,
             delegator_map[delegator] = {}
             delegator_map[delegator]['staking_balance'] = balance
             delegator_map[delegator]['current_balance'] = dexter_liquidity_provider_map[delegator]['current_balance']
+        delegator_map[delegator]['originaladdress'] = contract_id
```
jdsika, line 34: Is this by intention not in the "else" case? What is the commented code below. Can it be removed? If the code is a todo please remove and put it in an issue

amzid: I agree! If one delegator has XTZ directly delegated then the original fee ratio should apply not the one from the Dexter contract, thanks for the review! cleanup done of this file done!

jdsika (approved):

PR #307, opened 2020-11-15
Phase Reordering

utdrmac:
Consider the following config. tz1boot is baker, owner, and founder. 
```
baking_address: tz1boot1pK9h2BVGXdyvfQSv8kd1LQM6H889
founders_map:
  tz1boot1pK9h2BVGXdyvfQSv8kd1LQM6H889: 0.2
  KT1KLQbYFtFZ5mAEnfEMZaWYuNtCsGuP5cLS: 0.8
owners_map:
  tz1boot1pK9h2BVGXdyvfQSv8kd1LQM6H889: 0.3
  KT1VyxJWhe9oz3v4qwTp2U6Rb17ocHGpJmW0: 0.7
rules_map:
  tz1boot1pK9h2BVGXdyvfQSv8kd1LQM6H889: TOB
```

During the phase0-4 calculations, 3 reward records will be created: 1 for being a founder, 1 for being owner, and 1 for being self-delegator. Right now, the phase which maps addresses to different destinations happens _before_ the payment merging happens.

During the merge phase, the payout record for the founder and owner reward will overwrite the 'TOB' destination and TRD will create a transaction from tz1boot -> tz1boot.

This PR renames phases 5, 6, and 7 to something more appropriate and easier to recognize. This PR also changes when remapping of rules_map takes place. This now happens _after_ merge phase.

* **Check list**:

[NA] I extended the Travis CI test units with the corresponding tests for this new feature (if needed).
[NA] I extended the Sphinx documentation (if needed).
[NA] I extended the help section (if needed).
[NA] I changed the README file (if needed).
[NA] I created a new issue if there is further work left to be done (if needed).

**Work effort**: 8hrs

PR #359, opened 2021-01-24
Fix reordering bug causing double transactions to redirected payment addresses

amzid:
This PR reorders the mapping and merge phases to avoid double transactions.
The reordering introduced in #307 was motivated by an edge case, but it is causing double transactions for all redirected payments reported by some bakers since mid November.
@utdrmac I just reordered the phases again to their original order, I am not sure I understand the motivation of #307, would it be possible to find an alternative solution that does not cause double payments? 

**Work effort**: 0.5 hrs.

### DIFF
--- a/src/pay/payment_consumer.py
+++ b/src/pay/payment_consumer.py
@@ -82,13 +82,13 @@ class PaymentConsumer(threading.Thread):
 
                 logger.info("Starting payments for cycle {}".format(pymnt_cycle))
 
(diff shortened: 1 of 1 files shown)


utdrmac (approved):

utdrmac:
Hrm, I'll go back and try to find the original issue.

jdsika (approved):

Issue #365, opened 2021-02-03
Unable to parse config & Payment Address cannot be translated into a PKH


Twente-Mining:
When trying to start the payout software with the latest version an error arises.

```

Unable to parse '/home/ubuntu/hsmpay/tz.......yaml' config file.
MainThread - Payment Address (payout) cannot be translated into a PKH. Make sure it is a tz1 address and to first import its corresponding secret key to the signer.

```

amzid:
Are you using an alias?

Twente-Mining:
yes

amzid:
The use of an alias is not supported anymore sue to the removal of the dependency from the Tezos-client. Please use the pkh instead, setup the Tezos-signer and try again!

Twente-Mining:
ah yes, now it starts again. 
Thanks

jdsika:
@amzid can we introduce a check for aliases and return an error message that instructs for the use of the addresses instead?

amzid:
The check is happening and the error message instructs the user to use a tz1 address.

jdsika:
we could also allow the usage of aliases again and import them using the tezos-signer ? We intend to keep this dependency anyway, right?

amzid:
Right now there is no direct way to query the signer about any aliases. We have an issue in the Tezos Gitlab repo regarding the use of the signer. We can add this functionality as a feature request in the issue.

jdsika:
@utdrmac can you check the config for aliases and throw and error that instructs the user to exchange it with the address?

utdrmac:
```
This application will help you configure TRD to manage payouts for your bakery. Type enter to continue >
Specify your baking address public key hash (Processing may take a few seconds) >
adsf
Invalid baking address: Incorrect input in baking address, 'adsf' is not a tz or KT address
Specify your baking address public key hash (Processing may take a few seconds) >
tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7H
Invalid baking address: Incorrect input in baking address, 'tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7H' length must be 36
Specify your baking address public key hash (Processing may take a few seconds) >
tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7Hs
Specify your payouts public key hash. It can be the same as your baking address, or a different one. >
myalias
Invalid payouts address: Incorrect input in payouts address, 'myalias' is not a tz or KT address
[...]

PR #368, opened 2021-02-08
Add rewards type to configurator

utdrmac:
Adds asking the question of what payouts type do you want to use to configure.py

**Work Effort**: 8hrs (This was a lot of extra work, trying to figure out how to get python-yaml to dump a plain string vs some yaml-syntax for the custom Enum object. It was stupidly difficult in yaml; easy for json)

### DIFF
--- a/src/NetworkConfiguration.py
+++ b/src/NetworkConfiguration.py
@@ -54,6 +54,6 @@ def parse_constants(constants):
     network_config_map['BLOCK_TIME_IN_SEC'] = int(constants['time_between_blocks'][0])
     network_config_map['BLOCKS_PER_CYCLE'] = int(constants['blocks_per_cycle'])
     network_config_map['BLOCKS_PER_ROLL_SNAPSHOT'] = int(constants['blocks_per_roll_snapshot'])
-    network_config_map['BLOCK_REWARD'] = int(constants['baking_reward_per_endorsement'][0] * constants['endorsers_per_block'])
+    network_config_map['BLOCK_REWARD'] = int(int(constants['baking_reward_per_endorsement'][0]) * int(constants['endorsers_per_block']))
     network_config_map['ENDORSEMENT_REWARD'] = int(constants['endorsement_reward'][0])
     return network_config_map
--- a/src/configure.py
+++ b/src/configure.py
@@ -12,6 +12,7 @@ from fysom import Fysom
 from NetworkConfiguration import init_network_config
 from api.provider_factory import ProviderFactory
 from cli.client_manager import ClientManager
+from Constants import RewardsType
 from config.config_parser import ConfigParser
 from config.yaml_baking_conf_parser import BakingYamlConfParser
 from config.yaml_conf_parser import YamlConfParser
@@ -21,7 +22,7 @@ from launch_common import print_banner, add_argument_network, add_argument_repor
 from log_config import main_logger, init
 from model.baking_conf import BakingConf, BAKING_ADDRESS, PAYMENT_ADDRESS, SERVICE_FEE, FOUNDERS_MAP, OWNERS_MAP, \
     MIN_DELEGATION_AMT, RULES_MAP, MIN_DELEGATION_KEY, DELEGATOR_PAYS_XFER_FEE, DELEGATOR_PAYS_RA_FEE, \
-    REACTIVATE_ZEROED, SPECIALS_MAP, SUPPORTERS_SET
+    REACTIVATE_ZEROED, SPECIALS_MAP, SUPPORTERS_SET, REWARDS_TYPE
 from util.address_validator import AddressValidator
 from util.dir_utils import get_payment_root, get_successful_payments_dir, get_failed_payments_dir
 from util.fee_validator import FeeValidator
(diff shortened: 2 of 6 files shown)


utdrmac:
Also cleaned up configure.py a bit. Removed some un-used functions and corrected some wording.

utdrmac:
Resolves #365 
```
This application will help you configure TRD to manage payouts for your bakery. Type enter to continue >
Specify your baking address public key hash (Processing may take a few seconds) >
adsf
Invalid baking address: Incorrect input in baking address, 'adsf' is not a tz or KT address
Specify your baking address public key hash (Processing may take a few seconds) >
tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7H
Invalid baking address: Incorrect input in baking address, 'tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7H' length must be 36
Specify your baking address public key hash (Processing may take a few seconds) >
tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7Hs
Specify your payouts public key hash. It can be the same as your baking address, or a different one. >
myalias
Invalid payouts address: Incorrect input in payouts address, 'myalias' is not a tz or KT address
Specify your payouts public key hash. It can be the same as your baking address, or a different one. >
tz1RMmSzPSWPSSaKU193Voh4PosWSZx
Invalid payouts address: Incorrect input in payouts address, 'tz1RMmSzPSWPSSaKU193Voh4PosWSZx' length must be 36
Specify your payouts public key hash. It can be the same as your baking address, or a different one. >
tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7Hs
Specify bakery fee [0:100] >
33
```

jdsika:
Is an integrity check of the config running each time the config is loaded? Please integrate a check for correct values (like the percentage values - see slack error message).

utdrmac:
@jdsika Yes, each time in main.py the config is loaded, parsed, and validated. Need to see the bad config to understand what input the user has provided that is passing the checks.

jdsika:
@utdrmac do you check the rules/boundaries of the different fields? reward split needs to be a value between 0 and 1.. so on and so forth?

jdsika:
@habanoz please review

utdrmac:
@jdsika Yes, that code has been in place since the beginning. It checks that all founders/owners and other ratios total to 1.0

jdsika:
I tagged you at some question in the slack were someone had super strange payout amounts. Someone else said it might be related to using comma instead of dots for decimal separators?

utdrmac:
@jdsika Yes, I saw it, and saw the config the user posted. Looks perfectly normal.
I tested using `founders_map: { tz1RMmSzPSWPSSaKU193Voh4PosWSZx1C7Hs: 1,0 }`
and it throws errors at app startup where it is supposed to.

utdrmac:
Found the issue. Seems python doesn't always parse string as int when converting from json. The network constants were being interpreted as a string rather than int. Updated code to force int math.

Work Effort: 2hrs (debugging and testing fix)

src/NetworkConfiguration.py
```diff
@@ -54,6 +54,6 @@ def parse_constants(constants):
     network_config_map['BLOCK_TIME_IN_SEC'] = int(constants['time_between_blocks'][0])
     network_config_map['BLOCKS_PER_CYCLE'] = int(constants['blocks_per_cycle'])
     network_config_map['BLOCKS_PER_ROLL_SNAPSHOT'] = int(constants['blocks_per_roll_snapshot'])
-    network_config_map['BLOCK_REWARD'] = int(constants['baking_reward_per_endorsement'][0] * constants['endorsers_per_block'])
+    network_config_map['BLOCK_REWARD'] = int(int(constants['baking_reward_per_endorsement'][0]) * int(constants['endorsers_per_block']))
```
habanoz, line 57: Is the wrapper int really necessary?

habanoz: Not a blocker.

src/configure.py
```diff
@@ -33,8 +32,9 @@
 messages = {
     'hello': 'This application will help you configure TRD to manage payouts for your bakery. Type enter to continue',
     'bakingaddress': 'Specify your baking address public key hash (Processing may take a few seconds)',
-    'paymentaddress': 'Specify your payment PKH. Available addresses:{}',
+    'paymentaddress': 'Specify your payouts public key hash. It can be the same as your baking address, or a different one.',
```
habanoz, line 35: "Available addresses:" is removed. Could be useful for the user.

src/configure.py
```diff
@@ -97,8 +100,23 @@ def onservicefee(input):
...
+    if not input:
+        input = 'actual'
+
+    try:
+        global parser
+        rt = RewardsType(input.lower())
+        parser.set(REWARDS_TYPE, str(rt))
     except Exception:
-        printe("Invalid service fee: " + traceback.format_exc())
+        printe("Invalid option for rewards type. Please enter 'actual' or 'ideal'.")
```
habanoz, line 119: Considering the fact that there may be more reward type options in the future, Listing enum values would be more future proof. 
Since new reward types are not very likely, not a blocker.

src/NetworkConfiguration.py
utdrmac: The outer one? Probably not, but fits the pattern of all the other parameters so i left it.

habanoz (approved):
Nice job. Please see my comments.

habanoz:
@utdrmac Please see my comments.

habanoz:
@jdsika approved.

src/configure.py
utdrmac: It was removed because there was no code which was formatting the string for display. On my screen it literally showed "Available addresses:{}"

src/configure.py
jdsika: I agree and will keep this in mind.

Issue #372, opened 2021-02-14
Stateless injection waiting time


amzid:
**Describe the bug**
Stopping the TRD just after injecting the operation and before the operation is included is with the current implementation a critical action that might cause double payments after the TRD restarts.

**To Reproduce**
1. Start TRD and wait until op is injected then interrupt the process.
2. Restart the TRD after a few time and the cycle will be paid out again because no csv file was saved about any successful/failed/injected payment items. 

**Expected behavior**
A clear stepwise state for the TRD especially after the injection.

habanoz:
TRD must not be shut down while it is busy. TRD normally does not shut down immediately after sending ctrl+c.  Before termination, it waits to terminate active operations and prompts the user to wait until sensitive operations are completed. A second ctrl+c will terminate trd, which leaves the responsibility to the user.

jdsika:
The problem is the "unattended mode" in combination with server crashes. We need a proper state machine implemeted also as pre work for the database support @utdrmac . I am proposing a [Mealy-Machine](https://en.wikipedia.org/wiki/Mealy_machine)

jdsika:
[Mealy-Moore-Tutorial](https://www.tutorialspoint.com/automata_theory/moore_and_mealy_machines.htm)

jdsika:
[SQLite3 and Python](https://stackoverflow.com/questions/14509269/best-method-of-saving-data)

jdsika:
[MySQL and Python](https://www.w3schools.com/python/python_mysql_getstarted.asp)

utdrmac:
MySQL would not be a good choice as that requires running a "mysql server". SQLite is the far better choice. As soon as the operation is sent to the RPC, the RPC immediately returns the opHash which should be saved to DB immediately with an unknown state. Upon TRD restart, check the DB for any unknown opHash states and query RPC for state and update accordingly.

habanoz:
> As soon as the operation is sent to the RPC, the RPC immediately returns the opHash which should be saved to DB immediately with an unknown state. Upon TRD restart, check the DB for any unknown opHash states and query RPC for state and update accordingly.

I agree.

> MySQL would not be a good choice as that requires running a "mysql server"
I agree.

> SQLite is the far better choice.
@utdrmac Do you have sqlite production experience? How reliable is it? We need a system that assures no data is lost. I am worried, a system crash during a DB operation may leave the DB file corrupt? Is it a legitimate concern?

[...]

PR #373, opened 2021-02-14
Fix post-injection fetching of op hash

amzid:
This PR resolves an issue which was redundantly reported by some bakers. The following steps were performed:

* **Analysis**: 
  * issues reported that the op hash is marked as injected even though the block explorer shows applied status
  * Even more critical support requests: unclear timeout leading to interrupting process, restarting causes double payment (#372)
 
* **Solution**: More deterministic fetching of ops based on explicit levels and not waiting times, two trials per block, explicitly announced timeout!

* **Performed tests**: test Payouts in Delphinet.

* **Documentation**: no documentation required.

**Work effort**: 3 hrs including time which was needed for processing support requests.

### DIFF
--- a/src/pay/batch_payer.py
+++ b/src/pay/batch_payer.py
@@ -13,7 +13,8 @@ logger = main_logger
 
 MAX_TX_PER_BLOCK = 200
 PKH_LENGTH = 36
-PATIENCE = 10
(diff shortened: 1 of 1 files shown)


jdsika (approved):
@habanoz looks good to me

habanoz (approved):

habanoz:
Approved.

PR #384, opened 2021-02-28
Fix internal op parsing

amzid:


src/pay/batch_payer.py
```diff
@@ -315,12 +315,12 @@ def simulate_single_operation(self, payment_item, pymnt_amnt, branch, chain_id):
             # Calculate actual used storage
             storage = 0
             if 'paid_storage_size_diff' in op['metadata']['operation_result']:
-                storage += op['metadata']['operation_result']['paid_storage_size_diff']
+                storage += int(op['metadata']['operation_result']['paid_storage_size_diff'])
             if "internal_operation_results" in op["metadata"]:
                 internal_operation_results = op["metadata"]["internal_operation_results"]
```
jdsika, line 320: Why is there no cast here?

amzid: `op["metadata"]["internal_operation_results"]` is a dict, so I do not think that we need a cast here

jdsika (approved):

PR #386, opened 2021-03-03
Filter out skipped rewards

utdrmac:
* **Analysis**: Payment records which became skipped due to having zero balance were still included in the batch operation, causing the batch to backtrack/skip.
* **Solution**: Identify which payouts should be skipped and skip them.
* **Performed tests**: Tested against own bakery payouts for C328, with additional testing by @RTezos (TezosRUs)

**Work effort**: 4hrs (diagnosis and debugging with @RTezos, finalizing code changes)


utdrmac:
* updated the example config to include the correct telegram parameters.
* added useragent to webhook plugin because sometimes POST would be denied without it

habanoz (approved):

habanoz:
Fixed a minor logger formatting issue blocking Travis CI build.

jdsika (approved):

PR #387, opened 2021-03-16
Issue log gz

### LINKED ISSUE #385: Consider compressing verbose logs with gz
@habanoz Compressing verbose logs with zip requires manual decompression each time you want to view a log file. This is a painstaking process when you're trying to find a problem.

If the logs were compressed with gz, then native linux tools like 'zless', 'zgrep' can be used to read and search the logs without requiring a manual step to decompression.

habanoz:
This PR resolves #385 . 

Log archiving moved to gzip format from zip format.

**Work effort**: 0.5H

### DIFF
--- a/src/verbose_logging_helper.py
+++ b/src/verbose_logging_helper.py
@@ -2,7 +2,7 @@ import gzip
 import logging
 import os
 from datetime import datetime
-
+import shutil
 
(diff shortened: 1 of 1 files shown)


jdsika:
@utdrmac does it work as expected for you?

utdrmac (approved):

Issue #404, opened 2021-03-31
Broken Unit Test


habanoz:
**Describe the bug**
A unit test is broken. Needs to be fixed.

**To Reproduce**
1. Go to test/test_tzkt_api.py
2. Rename disabled_test_dry_run to test_dry_run
3. Run the test

**Expected behavior**
The test seems to be important. It checks details about payment calculations.

**Screenshots**
![image](https://user-images.githubusercontent.com/6825980/113202476-76fb3100-9273-11eb-89e2-a90b4b2a06a8.png)

jdsika:
I am closing this as I think it has been solved. @vkresch reopen if it was a mistake

PR #405, opened 2021-04-04
Unit tests refactoring

vkresch:
---
name: Test refactoring
about: Separated tests into meaningful classes and added 10 tests to the CI. 8 new tests and 2 previously skipped tests.
labels: 

---
* **Analysis**: Unit tests are currently complex and still use the builtin unit test library although it's ported to pytest.

* **Solution**: Simplify unit test via restructuring and using pytest syntax.

* **Implementation**: Added reusable classes and pytest structure. Furthermore the test are now divided into test class folders.

* **Performed tests**: Ran `pytest tests/` which gave the same result as the test runs on master. (Refactoring)

* **Documentation**: Added readme.md into the test directory.
* 
* **Check list**:

- [x] I created a new issue if there is further work left to be done (if needed).

**Work effort**: 11h


vkresch:
The last commit fixes https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/404 . 

@habanoz I'm not sure how the tzkt api request should be handle if it's empty (204 response) and thus cannot be decoded by `response.json()`. In the [regression test](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/405/files#diff-9b3bbcf05c3225f3093d03f4a663c4a6256c83556aeea2657bcadc289ba97b80R4) I just check if the API error is thrown after disabling 204 response to pass (see [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/405/files#diff-8ae56d2343982bb97105d3cb06ec24ca001ad4fb049d09747041517112989a2bR69)) . Should the API error even be thrown?

habanoz:
> @habanoz I'm not sure how the tzkt api request should be handle if it's empty (204 response) and thus cannot be decoded by `response.json()`. In the [regression test](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/405/files#diff-9b3bbcf05c3225f3093d03f4a663c4a6256c83556aeea2657bcadc289ba97b80R4) I just check if the API error is thrown after disabling 204 response to pass (see [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/405/files#diff-8ae56d2343982bb97105d3cb06ec24ca001ad4fb049d09747041517112989a2bR69)) . Should the API error even be thrown?

I am not sure if I understand your question correctly, I will try to provide an answer. 

From tzakt_api perspective;

I believe an empty response is not something we expect. I believe an empty response is a sign of failure. 
However, since 204 is explicitly added I believe the author felt the need to allow empty responses. 

Assuming an empty response is legal, then the caller must validate the response. Then we should not raise an API error and make sure the caller validates the response everywhere.

The alternative approach is assuming 204 is a failure and raising an API error. This one is easier but involves a little bit of risk.  @amzid @utdrmac colleagues may like to comment here.

From the application perspective;

An illegal baking address should be rejected during the configuration validation phase. I created #410.

src/tzkt/tzkt_api.py
```diff
@@ -65,7 +66,7 @@ def _request(self, path, **params):
         except requests.RequestException as e:
             raise TzKTApiError(e)
 
-        if response.status_code not in [200, 204]:
+        if response.status_code != HTTPStatus.OK:
```
jdsika, line 69: "ok" corresponds to two states? 200 and 204?

vkresch: It's only 200. `HTTPStatus.NO_CONTENT == 204`. see [docs](https://docs.python.org/3/library/http.html#http-status-codes).

src/log_config.py
```diff
@@ -5,7 +5,7 @@
 from verbose_logging_helper import VerboseLoggingHelper
 
 DEFAULT_LOG_FILE = 'logs/app.log'
-FORMATTER = logging.Formatter('%(asctime)s - %(threadName)-9s - %(message)s')
+FORMATTER = logging.Formatter('%(asctime)s - %(threadName)-9s - %(levelname)s - %(message)s')
```
vkresch, line 8: IMHO it would be interesting to see what kind of log message is currently displayed.

tests/integration/test_retry_producer.py
```diff
@@ -82,7 +80,7 @@ def test_retry_failed_payments(self):
         nb_success = len([row for row in success_report_rows if row.paid == PaymentStatus.PAID])
         nb_hash_xxx_op_hash = len([row for row in success_report_rows if row.hash == 'xxx_op_hash'])
 
-        self.assertEqual(31, nb_success)
+        self.assertEqual(26, nb_success)
```
vkresch, line 83: Just to make sure I understand it correctly:
There are total 31 batch payments. 5 are `nb_hash_xxx_op_hash` which are not successful thus making 26 sucessful payments right?

src/pay/batch_payer.py
```diff
@@ -503,10 +505,11 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
             status = -1
             list_op_hash = []
             trial_i = 0
-            while not (status == 200) and (trial_i < MAX_NUM_TRIALS_PER_BLOCK):
+            while status != HTTPStatus.OK and (trial_i < MAX_NUM_TRIALS_PER_BLOCK):
                 sleep(self.network_config['BLOCK_TIME_IN_SEC'])
                 status, list_op_hash = self.clnt_mngr.request_url(cmd)
-            if not (status == 200):
+                trial_i += 1
```
vkresch, line 511: Found an infinite loop.

tests/integration/test_retry_producer.py
habanoz: @vkresch This is a test about retrying failed operations. Input is a past payment report which contains 31 payment items, 26 of them were successful and 5 of them were failed. 

The test checks that all items are successful in the final report, so there must be 31 successful items.
And also checks that only the previously failed items are repaid. This is crucial, already paid items must not be paid again, so only 5 of them have the test-specific hash value. 

Even though the test configuration was correct, I admit it is not readable.

src/log_config.py
habanoz: @vkresch 

The app.log is for the user. The verbose log is for the maintainers. I think level information in the app.log is not very relevant.

src/pay/batch_payer.py
habanoz: Good catch ! :)

habanoz (commented):
Please see my comments!

tests/integration/test_retry_producer.py
vkresch: @habanoz shouldn't the final report have the `DONE` payment status for all 31 items instead of `PAID` and `INJECTED`?

vkresch:
Since I do not want to make this PR bigger I will open a separate PR for further refactoring (using explicitly pytest) or regression tests.

src/tzkt/tzkt_api.py
jdsika: yes, it is way more expressive but why did you remove then the `HTTPStatus.NO_CONTENT == 204` was 204 not correct or irrelevant?

src/log_config.py
jdsika: ok, leave the levelname and we see how users react on it.

tests/integration/test_retry_producer.py
habanoz: We have the following terminal states:
PAID: payment successfully made. 
FAIL: Some failures happened in the process.
DONE: Process completed without payment. E.g. zero amount, dry run...
INJECT: Transaction is injected into the node but after waiting for some time it is not added to any block. TRD does not know its fate. 

PAID and DONE means success. INJECT means unknown. 

In this particular test, we do not expect any payments with 'DONE' status.

src/tzkt/tzkt_api.py
vkresch: see [comment](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/405#issuecomment-813586548) above

tests/integration/test_retry_producer.py
```diff
@@ -81,11 +91,17 @@ def test_retry_failed_payments(self):
         self.assertTrue(os.path.isfile(success_report))
 
         success_report_rows = CsvPaymentFileParser().parse(success_report, 10)
-        nb_success = len([row for row in success_report_rows if row.paid == PaymentStatus.PAID])
-        nb_hash_xxx_op_hash = len([row for row in success_report_rows if row.hash == 'xxx_op_hash'])
+        success_count = len([row for row in success_report_rows])
```
vkresch, line 94: @habanoz I updated the test to just check the row count without the condition `PAID` since the csv report is within the `done` folder meaning the rows mark successful transactions. To be sure I count the file number in the failed directory and assert it to zero. Let me know if the test is fine like this.

tests/integration/test_retry_producer.py
vkresch: @habanoz  thanks for the clarification! I updated the unit test. See my comment further below.

tests/regression/test_simulate_single_operation.py
```diff
@@ -0,0 +1,65 @@
...
+        }
+    ]
+}
+
+
+@patch(
+    "cli.client_manager.ClientManager.request_url_post",
+    MagicMock(return_value=(HTTPStatus.OK, run_ops_parsed)),
+)
+def test_simulate_single_operation():
```
vkresch, line 30: @jdsika @utdrmac @amzid I added a regression test for the consumed_gas estimation here please update or add it according to the new PR (https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/411).

tests/regression/test_simulate_single_operation.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    reward_log.amount = 15577803
+    reward_log.skipped = False
+    simulation_status, simulation_results = batch_payer.simulate_single_operation(
+        reward_log, reward_log.amount, "hash", "unittest"
+    )
+    assert PaymentStatus.DONE == simulation_status
+    consumed_gas, tx_fee, storage = simulation_results
+    assert 150 == consumed_gas
+    assert 589 == default_fee + consumed_gas * MUTEZ_PER_GAS_UNIT
```
vkresch, line 64: Here is the current fee estimation.

jdsika (approved):

PR #411, opened 2021-04-07
Better Gas Estimations for Oven KT1

utdrmac:
This PR adds some additional checks, logs, etc for dealing with KT1 oven contracts.

**Work effort**: 6hrs


src/pay/batch_payer.py
```diff
@@ -336,9 +347,13 @@ def simulate_single_operation(self, payment_item, pymnt_amnt, branch, chain_id):
             logger.debug("Error while validating operation - Status: {}, Message: {}".format(status, op_error))
             return PaymentStatus.FAIL, []
 
+        consumed_gas += GAS_SAFETY  # Tezos client does this
+        consumed_storage += STORAGE_SAFETY
+
         # Calculate needed fee for extra consumed gas
-        tx_fee += int(consumed_gas * MUTEZ_PER_GAS_UNIT)
-        simulation_results = consumed_gas, tx_fee, storage
+        tx_fee += int((consumed_gas - int(self.gas_limit)) * MUTEZ_PER_GAS_UNIT)
```
amzid, line 354: I think we need to increase the fee also with some fee offset, many transactions are not being included currently

utdrmac: Should I remove the `-self.gas_limit` then? That would add a bit back to the fee.

keefertaylor: In the short term, this will work. I highly suspect that playing whack-a-mole with smart contracts isn't going to be efficient long term, and I think TRD would benefit from moving to an approach where estimates are done dynamically.  

There's a few things you can do to make this work. Any of the following is probably worth doing in isolation, but taken together they make up a pretty decent solution

- It's easy to estimate a gas and storage limit for an operation using a `pre-apply` (if you need help to get this working, LMK)
- Once you know fees / gas for each operation, you can batch them up according two two measures (i) max operations in a batch [1] and (ii) gas limit

```
const batches: Array<Array<operation>> = []
let currentBatch = 0
let currentGasUsedInBatch = 0
for (i = 0; i < operations.length; i++) {
  const operation = operations[i]

  // Init a new batch if needed
  if (batches[currentBatch] == undefined) {
    batches[currentBatch] = []
  }
  
  // Start a new batch if we've hit the max in a batch
  if (batches[i].length > OPS_PER_BATCH) {
    currentBatch += 1
    currentGasUsedInBatch = 0
  }
  
  // Start a new batch if we hit the gas limit
  if (currentGasUsedInBatch + operation.gas_limit > OP_GAS_LIMIT) {
    currentBatch += 1
    currentGasUsedInBatch = 0
  }
    
  batches[currentBatch].push(operation)
  currentGasUsedInBatch += operation.gas_limit 
}

```
(please excuse the fact that I don't speak python)

- Once you have batches of ops, you can estimate fees offline using static analysis. Here's an example implementation in JS: https://github.com/tacoinfra/harbinger-lib/blob/master/src/operation-fee-estimator.ts. You could optionally configure this to take a price per gas unit, which is probably a better measure than an offset since that is the metric bakers are going to look at. 


[1] In the short term, it probably makes sense to make this a configurable command line option, since I think other bakers may soon hit this problem and it's the easiest fix.

jdsika:
Todo: @vkresch added a regression test for the single operation simulation - please update or add it according to the this PR.

utdrmac:
@amzid I removed the safety. It is not needed when using `run_operation` it seems as that is the actual calculation. Using the safety resulted in inaccurate payouts, according to BB. The extra 100 gas + 20 storage results in 0.005 underpayment. By removing these two, payouts were accurate.

amzid:
The gas and storage calculations are accurate. The only thing that needs to be customly adjusted is the fee value. The current value is not always enough. We need to increase it by some safety margin (optionally adjustable by the user) or increase the factor mutez per gas unit. In some cases, the fees are not enough for inclusion and the batch transaction is lost in the mempool.

jdsika:
@utdrmac if you are ok with it I have asked @amzid to finish this PR even if it was originally your suggestion. I think it is valubale if he sortes the issue out as he was originally proposing this method. @vkresch will in the end design a test for it and I will invite you for review afterwards. I expect the PR to be ready on sunday to finally do a stable release next week. Those gas/fee calculations are a delicate thing ....

utdrmac:
@jdsika That's fine with me. Yes, I had several batches get lost in mempool. Recovering from that is very hard with TRD. Batches 1 and 4 went fine, but 2 and 3 were lost. TRD couldn't handle that situation.

src/pay/batch_payer.py
amzid: thank you for the comment keefer! the first step is already being performed correctly (gas+storage estimation), any failure related to gas/storage is detected and reported already at run_operation stage. Since problems are happening at the inclusion stage, the problem is most probably related to fee adjustment. 
image
Here you keep iterating till current fee is lower than required fee, I do not see however what is the changing parameter between two different iterations (head block?)

amzid:
> @jdsika That's fine with me. Yes, I had several batches get lost in mempool. Recovering from that is very hard with TRD. Batches 1 and 4 went fine, but 2 and 3 were lost. TRD couldn't handle that situation.

After comparing TRD code with Keefer's code, our fee estimation was only based on the consumed gas + base fee representing a fixed op bytes size, whereas in the code of Keefer it consists additionally of storage cost and variable bytes size cost. The storage cost was added by @utdrmac already, I guess this was causing many problems. Do you still get failing transactions (lost in mempool) even after these changes?

amzid:
Based on the feedback of @keefertaylor, I made two main changes:
* Batching: addresses are sorted according to their type, 200 is the batch size for tz1 addresses, 10 is the batch size for KT1 addresses.
* Better fee estimation: the size of the serialized batch operation is computed and included in the fee estimation

jdsika:
Did you test it on florence test net as well?

jdsika:
I am really happy about the sorting!!!

vkresch:
@amzid is this feature ready for a test? ;)

amzid:
@vkresch yes! I tested it on edo2net! Further tests on florence are needed (also mainnet to test the batching with some baker with a large number of delegators)

jdsika:
@amzid are the parameters configurable parameters now part of the baker configuration? Is fee.ini replaced?

amzid:
This was a fix for KT1 fee estimation. Tz1 addresses are still handeled according to fee.ini, I think a step-by-step approach is better. It is easier to test small changes on the mainnet than big ones.

src/pay/batch_payer.py
```diff
@@ -160,19 +166,31 @@ def pay(self, payment_items_in, dry_run=None):
             if pi.needs_activation and self.delegator_pays_ra_fee:
                 # Need to apply this fee to only those which need reactivation
                 zt += RA_BURN_FEE
-                sum_burn_fees += RA_BURN_FEE
 
+                # Check here if payout amount is greater than, or equal to new zero threshold with reactivation fee added.
+                # If so, add burn fee to global total. If not, payout will not get appended to list and therefor burn fee should not be added to global total.
+                if pi.amount >= zt:
```
vkresch, line 172: what is zt and pi? It would be good to spell out every variable for me who is new to the code base it's hard to read.

src/pay/batch_payer.py
```diff
@@ -160,19 +166,31 @@ def pay(self, payment_items_in, dry_run=None):
...
 
         if not payment_items:
             logger.info("No payment items found, returning...")
             return payment_items_in, 0, 0, 0
 
         # split payments into lists of MAX_TX_PER_BLOCK or less size
         # [list_of_size_MAX_TX_PER_BLOCK,list_of_size_MAX_TX_PER_BLOCK,list_of_size_MAX_TX_PER_BLOCK,...]
-        payment_items_chunks = [payment_items[i:i + MAX_TX_PER_BLOCK] for i in
-                                range(0, len(payment_items), MAX_TX_PER_BLOCK)]
+        payment_items_tz = [payment_item for payment_item in payment_items if payment_item.paymentaddress.startswith('tz')]
```
vkresch, line 187: It would be nice to have this as a function `split_payments()` so that it can be also unit tested.

src/pay/batch_payer.py
```diff
@@ -443,6 +484,25 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
...
+            total_fees = required_fee
+            content_list[0] = json.dumps(first_tx)
+            contents_string = ",".join(content_list)
+            forge_json = FORGE_JSON.replace('%BRANCH%', branch).replace("%CONTENT%", contents_string)
+            status, bytes = self.clnt_mngr.request_url_post(self.comm_forge, forge_json)
+            if status != HTTPStatus.OK:
+                logger.error("Error in forge operation")
+                return PaymentStatus.FAIL, ""
+            size = SIGNATURE_BYTES_SIZE + len(bytes) / 2
+            required_fee = math.ceil(MINIMUM_FEE_MUTEZ + MUTEZ_PER_GAS_UNIT * total_gas + MUTEZ_PER_BYTE * size)
```
vkresch, line 504: Is it possible to reuse this as a function with total gas and size as variables? Because it's already used above. It would be nice to have this block as a function to also test the output and have a comment what this block is doing.

vkresch (commented):
I added a high level unit test for the `pay` function of batch_payer if it calculates the total payment  correctly and added some minor comments. 
Not part of this PR but it would be good to have the function `pay` and `attempt_single_batch` split up into smaller function so that they can be tested more easily.

amzid:
Thanks for the input @vkresch ! Based on the discussion, I think we could merge this PR and do the refactoring together with the test cases in a separate PR!

jdsika (approved):

mkaczanowski:
@utdrmac @vkresch why max batch size for KT1 addresses is so small? This change significantly increased amount of batches to be executed in my case.

I am wondering if I can bump it up and to what higher number, if at all?

jdsika:
@mkaczanowski you could provide some systematic testing on that and come up with an empirical number or (much better) try to help derive a Formular to be used here

PR #415, opened 2021-04-11
introduce "self-ideal" rewards like Backerei

nicolasochem:
I am the maintainer of backerei and in the process of depcrecating it.

Backerei payout operations are different than TRD: its default,
non-configurable behaviour is to pay rewards assuming perfect behaviour
of the baker doing the payouts, while still accounting for imperfect
behaviour of other bakers. This acts as a liveness guarantee and ensures
maximum payouts for delegators, but without overpaying for other baker's
downtime, like the current "ideal/expected" payout does.

This PR adds this capability to TRD.

Nomenclature
============

The current master branch has "ideal" and "actual" payouts. I initially
thought about naming this new payout type "ideal_self" or "self-ideal"
but it makes much more sense to do some renaming:

| old  | new    |
| ---- | ------ |
|ideal |~expected~ estimated|
|actual|actual  |
|      |ideal   |

The new names are more self-explanatory: "pay estimated rewards" convey
that these rewards are not realized yet, while "pay ideal rewards"
includes the fees and extra rewards, which makes sense.

A drawback is that this will require a configuration change to current
users and will need to be release-noted. But, we expect most users of
the old-style "ideal/expected" rewards to be using them to pay out
future cycles. This is not possible with the new style "ideal" type,
because this reward type expect the cycle to actually have run, so these
users will have an error upon upgrade, which will force them to upgrade
their configuration from `rewards_type: ideal` to `rewards_type:
expected`.

I am open to any suggestions about better naming for this new model.

Implementation
==============

~It only works with tzstats and tzkt API; at the moment, the TRD
implementation relying on Tezos Node RPC does not implement detailed
calculation of rewards, instead relying on unfreezing events to
determine the rewards. So, there is no way to add this feature without
implementing the detailed calculation itself. As a result, this type of
rewards will error out when attempted to run with a Tezos Node RPC
backend.~

This feature was implemented in all 3 payment methods (tzkt, tzstats, rpc).

Both tzkt and tzstats "rewards" endpoint provide fields for missed
baking and endorsing rewards. When rewards_type.isIdeal is true, we add
these rewards to the "actual" rewards calculation, and this becomes our
total rewards.

~While doing this, I observed that TRD was never paying out stolen block
rewards when used with the tzstats backend. This bug has been fixed.~

I also observed that tzkt's rewards endpoints (but not tzstats) exposes
these fields:

* `missedOwnBlockFees`
* `missedExtraBlockFees`
* `uncoveredOwnBlockFees`
* `uncoveredExtraBlockFees`

These are the fees that the baker would have earned, had they baked the
blocks that they missed. We include them in the calculation because it
makes sense to do so. tzstats does not expose these lost fees, so there
is an ideal reward discrepancy between tzkt and tzstats backend when the
baker fails to bake one or more blocks in the cycle, with tzstats
backend reporting ideal payout lower than its true value. But this is
an edge case so it makes sense to enable ideal payouts with tzstats
backend anyway. We documented it.

For RPC we scan blocks with baking and endorsing rights and check whether
baking/endorsing happened. When it didn't, we add 40/1.25 tez to the rewards.


utdrmac:
If a payout style cannot be calculated using all three available methods, then it shouldn't be implemented.

I'm having difficulty understanding your change in payout calculations.

you said:
> pay rewards assuming perfect behaviour of the baker doing the payouts

But that's exactly the definition of ideal rewards: bake every prio 0 block and endorse every right. This is the "ideal" scenario. With ideal, if you miss a bake or an endorsement, you still pay out as if you did submit that bake.

Obviously, actual rewards doesn't include those bakes you missed.

Can you please rephrase what your new payouts term means?

nicolasochem:
Hi Matthew, thanks for looking. PR is not ready but description is accurate.

Say you baked and endorsed perfectly during one cycle, but due to some other baker being offline, you endorsed a priority 1 block due to no fault of your own. You are not going to earn the full 1.25 tez for this block, but if you pay expected rewards, you will pay out delegators based on the full amount. Same thing if you bake a block that does not contain enough endorsements because other bakers were offline. Because of this, you end up overpaying.

Using expected rewards, a large enough baker will almost certainly overpay at every cycle. The expected payouts option is thus not very useful, unless you pay before the cycle actually runs. In this case, you have no choice but to make predictions based on rights.

How the new "ideal" payout type is being calculated is: actual rewards + 1.25 per missed endorsement + 40 per missed block.

Another way of looking at the proposed "ideal" payout type: it pays the precise amount needed for baking bad to consider your payouts "accurate".

> If a payout style cannot be calculated using all three available methods, then it shouldn't be implemented.

Taking a closer look at the RPC code, it looks like I will be able to compute proposed "ideal" rewards with RPC as well (because it is possible to query the number of bakes and endorses per cycle), so if you agree in the merits of the feature, I'll go ahead and add it to this PR.

utdrmac:
> due to some other baker being offline, you endorsed a priority 1 block due to no fault of your own. You are not going to earn the full 1.25 tez for this block, but if you pay expected rewards, you will pay out delegators based on the full amount

That is not correct. If you bake a prio 1 TRD does not take that into consideration for 'ideal' payouts. Only prio0 bakes and all endorsements are considered for 'ideal'. Under 'actual', yes, you would pay additional rewards because you _actually_ earned extra rewards.

> Using expected rewards, a large enough baker will almost certainly overpay at every cycle.

How? If I'm a "large" baker and I bake 300 prio0 blocks in cycle 4 and I endorse 4000 times, both 'ideal' and 'actual' rewards are the same. Any baker, no matter "large" or "small", that has 100% efficiency in a cycle, ideal and actual rewards are the same.

We use 'actual' and BB shows accurate payouts for our delegators, so I'm not understanding this reasoning when it's already correct.

I'm not trying to be difficult, I'm just not understanding why we need to change something that is not broken. I'm not understanding why you think the current 'actual' and 'ideal' are calculated incorrectly. We have hundreds of bakers that use TRD every cycle and nobody has said anything about the inaccuracy of the calculated payouts.

docs/configuration.rst
```diff
@@ -53,10 +53,11 @@ Available configuration parameters are:
...
+  'expected': Rewards are calculated using the number of baking rights granted at priority 0, plus the count of all endorsing slots. If a bake or endorsement is missed, rewards are calculated as if there was no miss. No additional block rewards or transaction fees are included in this method. This allows to pay future rewards, before the cycle actually runs.
+  'ideal': Rewards are calculated assuming ideal behaviour of the baker, and actual behaviour of other bakers. If a bake or endorsement is missed, rewards are paid out despite not having been earned. Lost income due to other baker downtime are not included. Transaction fees and other block rewards are included in the rewards. Any lost rewards due to double-baking, double endorsing, or missing nonce are paid out anyway. Select this type of rewards to insure against downtime of your baker but still account for real world conditions. This way, you will get optimal ranking in baker evaluation services despite any downtime.
```
utdrmac, line 60: > Lost income due to other baker downtime are not included.

What do other bakers being offline have to do with your rewards?

nicolasochem:
>That is not correct. If you bake a prio 1 TRD does not take that into consideration for 'ideal' payouts. Only prio0 bakes and all endorsements are considered for 'ideal'.

Your statement is correct, but it is not what I said. If you endorse a priority 1 block, you will earn 0.833333 tez, [see documentation](https://tezos.gitlab.io/008/proof_of_stake.html#rewards). If you bake a prio 0 block but one of the endorsers of the previous block was offline, you will earn 38.75 tez.

> If I'm a "large" baker and I bake 300 prio0 blocks in cycle 4 and I endorse 4000 times, both 'ideal' and 'actual' rewards are the same. Any baker, no matter "large" or "small", that has 100% efficiency in a cycle, ideal and actual rewards are the same.

This is not correct for the reason stated above.

> We use 'actual' and BB shows accurate payouts for our delegators, so I'm not understanding this reasoning when it's already correct.

Provided you have 100% reliability, "actual" will show accurate payouts on BB. But if you fail to endorse a block or fail to block at prio 0, "actual" will show as too low on BB. With my new payout model, you will show as accurate and shoulder the cost of your downtime without passing it to your delegators.

> I'm not trying to be difficult, I'm just not understanding why we need to change something that is not broken. I'm not understanding why you think the current 'actual' and 'ideal' are calculated incorrectly.

Current "actual" and "ideal" payouts are calculated correctly. I am not changing any of the payout model, I am just introducing a third one which maps backerei's default behaviour.

Separately, I am proposing a renaming as follows:


old | new
-- | --
ideal | expected
actual | actual
  | ideal

docs/configuration.rst
nicolasochem: Addressed in main comment.

nicolasochem:
@utdrmac I have now added support for the rpc method as well.

Next I will add tests, but what is your sentiment? Do you understand where I am coming from with this new payout type?

utdrmac:
This is one of my oldest delegators, https://baking-bad.org/tz1RJTLee1JMDskpixR4mqBsT8iXGdZJJhX4, aside from our server migration in C335, he shows perfect payouts for past 50+ cycles. Surely, somewhere in all that time, we have missed endorsements, or the previous-block had 1 less endorsement. We've (and TRD) always used 'actual' payouts. Just to clarify, you don't see any issue with 'actual' correct?

And I think I understand what you're saying: The issue is that current 'ideal' assumes perfect behavior for you _and_ everyone else. Going on your example, if you bake block X prio 0, but block X - 1 only had 31 endorsements, instead of getting 40xtz, you get 38.75. TRD does not account for this. TRD assumes all rewards are earned under 'ideal' and thus you would be paying out an extra 1.25 that cycle because the baker of X-1 messed up, or an endorser didn't submit. Thus, you are being penalized for no fault of your own.

After thinking about this, I'm more inclined to merge the behavior you are proposing into 'ideal', rather than introduce a new behavior. I'm curious on @dansan566 and @jdsika opinions here.

nicolasochem:
> This is one of my oldest delegators, https://baking-bad.org/tz1RJTLee1JMDskpixR4mqBsT8iXGdZJJhX4, aside from our server migration in C335, he shows perfect payouts for past 50+ cycles. Surely, somewhere in all that time, we have missed endorsements, or the previous-block had 1 less endorsement. We've (and TRD) always used 'actual' payouts. Just to clarify, you don't see any issue with 'actual' correct?

No, I don't see any issue with actual.

It looks like I am wrong with TRD and it is "forgiving" you for missed endorsements. But the calculation method I am proposing here has been used for calculating Cryptium Labs payouts since cycle 11 (one of the largest bakers until recently). I think it has merit.

> After thinking about this, I'm more inclined to merge the behavior you are proposing into 'ideal', rather than introduce a new behavior. I'm curious on @dansan566 and @jdsika opinions here.

The issue is that you can not use proposed "ideal" payouts before the cycle runs (`RELEASE_OVERRIDE` of -11 to -6). But you can use existing "ideal". That's why I am suggesting to keep it and rename it to "expected".

jdsika:
Dear all,

thank you very much for the discussion and suggestions. I would very much take a "deep dive" into the topic today and discuss it with Daniel. I think we have to formulate some suiting definitions here that satisfy everyone and I am sure this is achievable. "Names" are amoungst the hardest things in programming :)

Best regards
Carlo

jdsika:
I will try to recap to see if I understood it correctly.

What does "actual" rewards mean for TRD:
* Devide and payout all rewards from all successful validation operations (baking/endorsing/nounce revelation). 
* Take into account that blocks get different rewards depending on the number of endorsements or their priority, as well as lower rewards for endorsing prio 1 (and lower) blocks.
* Clients gain with every additionally stolen block (prio 1) and they lose rewards with every missed bake and endorsement.
* Clients also participate in the block fees.
* Clients do NOT participate in the loss of frozen funds from slashing.
* Clients do NOT gain from slashing other bakers.
* Clients are not compensated for slashed rewards - it is an "unsuccessful validation operation".

What is "ideal":
* Clients get the full amount of rewards for a cycle as it has been "predicted" in the lottery.
* TRD has the assumtion that the network conditions will be perfect - so all blocks are actually baked in that cycle and they have all the endorsements.
* The clients acknowledges that the fees attached to blocks are NOT part of the payout.
* **@utdrmac I am unsure about Nounce revelation here.**
* Slashing/losing frozen rewards is NOT relevant for the client.

-> The naming is derived from the assumption that we will encounter "ideal" network conditions and a 100% participation of all in the lottery drawn validators.

What is "ideal" for baeckerei?
* It deviates from the TRD "ideal" definition as the assuption of a perfect behaviour is only made regarding your own baker.
* The failure of others (missed endorsement) resulting in lower block rewards is forwarded to the client resulting in lower payouts.


@nicolasochem I take from your conversation that you did NOT find any bug in the calculation method with regard to my explanation above?


I think the word "actual rewards" does refer correctly to what a baker does "actually" earn from successful validation operations and does share this with the clients. Slashed funds and gains from accusations are a both not accounted for as they are somehow both sides of the same medal.

The word "ideal" can imo refer correctly to an ideal network behaviour and at the same time it is also "expected" for people who are rather positive thinkers but I think the expectation is that the network does always have hicups. So I myself expect lost endorsements (we are often losing due to network errors or bugs in the software and not due to our own wrong doing). So for me it IS "ideal" and not "expected". 
I would rather say that what "cryptium labs" has been paying is somewhat expected.
* The expectation from the client that the baker does perform 100% (and that he is responsible for the software he is using)
* The expectation from the client that others they did not chose as their partner are not performing a 100%

nicolasochem:
> What does "actual" rewards mean for TRD:

> Clients gain with every additionally stolen block (prio 1) and they lose rewards with every missed bake and endorsement.

In the tzkt case, client do NOT gain with additional stolen block (prio 1). I fixed it in this PR.

> Clients do NOT gain from slashing other bakers.

Based on my reading of the code, clients DO gain from slashing other bakers. Please correct me if I'm wrong. 

> What is "ideal" for baeckerei?

> It deviates from the TRD "ideal" definition as the assuption of a perfect behaviour is only made regarding your own baker.
The failure of others (missed endorsement) resulting in lower block rewards is forwarded to the client resulting in lower payouts.

No, "ideal" for backerei does pay for fees, nonce revelations, stolen blocks and equivocation accusation reward. It deviates from the TRD "actual" definition by adding 1.25 tez per missed endorsement and 40 tez per missed bake to the total amount to be distributed.

> @nicolasochem I take from your conversation that you did NOT find any bug in the calculation method with regard to my explanation above?

I did find a bug where tzkt didn't pay for stolen blocks. I also think that TRD distributes the accusation rewards. (why wouldn't it?)

nicolasochem:
Thinking a bit more about this, maybe "pay estimated rewards" is the better way to put it.


old | new
-- | --
ideal | estimated
actual | actual
  | ideal

jdsika:
Let's try it with a table with definitions taken from BakingBad. An "X" indicates that the client participates, "-" not and "O" for optional depending on baker config.

Reward Types | Estimated | Actual | Ideal
-- | -- | -- | --
Block Rewards | X | X | X
Missed Blocks | - | X | -
Stolen Blocks | - | X | X 
Endorsement Rewards  | X | X | X 
Missed Endorsements | - | X | -
Low-priority Endorsements | - | X | X
Gained from Fees | - | X | X
Accusation Rewards | - | - | X
Accusation: Lost Deposit | - | - | -
Accusation: Lost Reward | - | X | -
Accusation: Lost Fees | - | X | X
Revelation Rewards | X | X | X
Revelation: Lost Rewards | - | X | -
Revelation: Lost Fees | - | X | X
Allocation Fee | O | O | O
Transaction Fee | O | O | O

nicolasochem:
Your table is both additive and substractive. May I suggest another table with only positive quantities. X means "taken into account for reward calculation" and "-" means "not taken into account".

Reward Types | Estimated | Actual | Ideal
-- | -- | -- | --
Baked (Prio 0) Blocks | X | X | X
Missed Blocks | X | - | X
Stolen (Prio >0) Blocks | - | X | X 
Successful Endorsements  | X | X | X 
Missed Endorsements | X | - | X
Baking Fees | - | X | X
Accusation Rewards | - | X | X
Revelation Rewards | - | X | X

I am also removing the transfer/allocation fees since I believe it's irrelevant for this stage where we calculate total rewards.

I also think the losses due to slashings and revelation failures are irrelevant in the total reward calculation. These losses are most certainly higher than any rewards, so you would not even be running a payout engine in such situation.

> If the code for "Actual" is paying out the accusation rewards it is in fact a bug in my opinion. The rationale is very clear for me:
The baker takes the risk of losing a deposit and he gains from accusing. This is two sides of a medal.

If you keep the accusation rewards, you may face pressure to transfer them back to the faulty party. If you distributed them to your delegates, then you can wash your hands of it and be immune from pressure.

We can make it configurable, but most people will stick to the default and I think that the default should be to pay them out. But this is a separate discussion and I'll fix the PR to not change today's behavior.

dansan566:
@nicolasochem I don't agree with you on the accusation reward. Often it is a mistake (start backup node but the main node is still running) and therefore human error and not an attack that leads to a double baking event. Often it is even a friendly baker. We would rather send the accusation reward back to the baker instead of distributing it. If this really happens, I think it should always be discussed and distributed (to delegators, slashed baker, keep it) separately.

jdsika:
The best option would be to drop the "categories" of "ideal" etc and just add the configuration table to the config file.

nicolasochem:
OK, I see the point. I agree that distributing the accusation rewards is controversial and it should be a config option.

I don't think everything should be a config option. Nothing else is controversial or ambiguous, and too many knobs is not always good, and "estimated/ideal/actual" is quite self-explanatory.

How about a toggle "pay accusation rewards" (default to no) and keep the three categories?

nicolasochem:
@jdsika @utdrmac @dansan566 it's ready for review.

Per our discussion, I have removed the payout for double baking/endorsement accusation on tzstats. It was not present in tzkt and rpc. So no accusation rewards will ever be distributed to the delegators, as is the case today on master branch. (this could be enabled with a specific configuration toggle later)

I have also updated the rpc code to accomodate a change in rpc after florence.

Since it has been a while, I ran a battery of tests again. I can confirm that it all works as described in the PR description.

dansan566:
@nicolasochem I did some rudimentary sample testing with tzstats and tzkt as providers. I tested three different cycles (377, 268 and 208) and wrote down the "total amount to pay **out**" number from the output. This sums up to 3*3*2 = 18 tests. I've chosen the cycles because of endorsement miss, block miss, and nonce reward. I've noticed that for cycles 268 and 208, tzstats gives me higher numbers compared to tzkt when choosing actual and ideal. Estimated is the same for both providers and both cycles. When I switched back to the master branch, I've recognized that the "total amount to pay out" is the same for actual for tzstats and tzkt!

https://airtable.com/invite/l?inviteId=invTVGzEyebvnyyoH&inviteToken=939d76513d634125eeef86615e4d626a388bcf72ac6c9e08151374575b5fd99e&utm_source=email

nicolasochem:
@dansan566 thank you for testing!

See from my PR discussion:

> While doing this, I observed that TRD was never paying out stolen block
rewards when used with the tzstats backend. This bug has been fixed.

From tzstats webpage I see "rescued rewards" of 6 tez for stakenow for cycle 268 and 10.6875 tez for cycle 208. This explains the difference that you are seeing. Rescued rewards are the rewards you get from baking blocks at priority 1 or more. These ought to be paid to delegates, but they are not today when you are using the tzstats api.

I stumbled upon this problem while doing changes on the tzstats api, but it looks like it's also present in tzkt. Consequently, I will not fix this as part of this PR and open a ticket instead.

dansan566:
@nicolasochem I think this is more of an API definition issue between both providers:
tzstats:
Bildschirmfoto 2021-07-26 um 09 49 45
tzkt:
Bildschirmfoto 2021-07-26 um 09 50 20
If you add the tzkt numbers (Own blocks + Extra blocks + Endorsements + Fees (Own Block + Extra Block) = 1,518,747348 which is the actual reward. tzstats shows exactly the same number in total income (1,518.747348). Therefore, I can't follow this statement: "These ought to be paid to delegates, but they are not today when you are using the tzstats API". Am I missing something here?

jdsika:
@nicolasochem I think in tzstats the rewards of prio 1 and higher blocks are already included in `(float(resp[idx_income_baking_income]) + float(resp[idx_income_endorsing_income])`

src/pay/payment_consumer.py
```diff
@@ -237,7 +237,12 @@ def create_stats_dict(self, nb_failed, nb_unknown, payment_cycle, payment_logs,
         stats_dict['nb_delegators'] = n_d_type
         stats_dict['pay_xfer_fee'] = 1 if self.delegator_pays_xfer_fee else 0
         stats_dict['pay_ra_fee'] = 1 if self.delegator_pays_ra_fee else 0
-        stats_dict['rewards_type'] = "I" if self.rewards_type.isIdeal() else "A"
+        if self.rewards_type.isIdeal():
+            stats_dict['rewards_type'] = "I"
+        elif self.rewards_type.isActual():
```
jdsika, line 242: @nicolasochem don't we need an `else:  stats_dict['rewards_type'] = "A"` at the end as the "default value"?

src/pay/payment_producer.py
```diff
@@ -250,7 +250,7 @@ def run(self):
 
         return
 
-    def try_to_pay(self, pymnt_cycle, expected_rewards=False):
```
jdsika, line 253: For the python noob. The variable `expected_rewards` was only defined as part of the parameter in the `try_to_pay` method and you removed it by referencing to the `self.reward_type` ?

Kind of horrible that an additional boolean has been introduced instead of using the reward types from the constants. I am glad this has been removed

src/rpc/rpc_reward_api.py
```diff
@@ -58,12 +60,12 @@ def get_rewards_for_cycle_map(self, cycle, expected_rewards=False):
                          .format(cycle, self.preserved_cycles, self.blocks_per_cycle,
                                  level_of_first_block_in_preserved_cycles, level_of_last_block_in_unfreeze_cycle))
 
-            # Decide on if paying actual rewards earned, or paying expected/ideal rewards
-            if expected_rewards:
```
jdsika, line 62: Is it correct that the naming in the code was inconsistent anyway? `expected_rewards` was used instead of the word `ideal_rewards` as the actual name derived from the configuration?

src/rpc/rpc_reward_api.py
```diff
@@ -58,12 +60,12 @@ def get_rewards_for_cycle_map(self, cycle, expected_rewards=False):
...
+            # Decide on if paying actual rewards earned, or paying estimated/ideal rewards
+            if rewards_type.isEstimated():
 
                 # Determine how many priority 0 baking rights delegate had
-                nb_blocks = self.__get_number_of_baking_rights(cycle, level_of_first_block_in_preserved_cycles)
-                nb_endorsements = self.__get_number_of_endorsement_rights(cycle, level_of_first_block_in_preserved_cycles)
+                nb_blocks = len([r for r in self.__get_baking_rights(cycle, level_of_first_block_in_preserved_cycles) if r["priority"] == 0])
```
jdsika, line 67: Why did you change this?

src/rpc/rpc_reward_api.py
```diff
@@ -73,7 +75,7 @@ def get_rewards_for_cycle_map(self, cycle, expected_rewards=False):
                 total_block_reward = nb_blocks * self.block_reward
                 total_endorsement_reward = nb_endorsements * self.endorsement_reward
 
-                logger.info("Ideal rewards for cycle {:d}, {:,} block rewards ({:d} blocks), {:,} endorsing rewards ({:d} slots)".format(
+                logger.info("Estimated rewards for cycle {:d}, {:,} block rewards ({:d} blocks), {:,} endorsing rewards ({:d} slots)".format(
                             cycle, total_block_reward, nb_blocks, total_endorsement_reward, nb_endorsements))
 
                 reward_data["total_rewards"] = total_block_reward + total_endorsement_reward
```
jdsika, line 81: I am confused here. This whole code block is inside the `if rewards_type.isEstimated():` ???

src/rpc/rpc_reward_api.py
```diff
@@ -83,10 +85,28 @@ def get_rewards_for_cycle_map(self, cycle, expected_rewards=False):
...
+                                missed_baking_income += self.block_reward
+                    missed_endorsing_income = 0
+                    for r in self.__get_endorsement_rights(cycle, level_of_first_block_in_preserved_cycles):
+                        authored_endorsement_slots = self.__get_authored_endorsement_slots_by_level(r["level"] + 1)
+                        if authored_endorsement_slots != r["slots"]:
+                            mutez_to_add = self.endorsement_reward * len(r["slots"])
+                            logger.warning("Found {} missed endorsement(s) at level {}, adding {} mutez reward anyway.".format(len(r["slots"]), r["level"], mutez_to_add))
+                            missed_endorsing_income += mutez_to_add
+                    logger.warning("total rewards %s" % (total_actual_rewards + missed_baking_income + missed_endorsing_income))
+                    reward_data["total_rewards"] = total_actual_rewards + missed_baking_income + missed_endorsing_income
```
jdsika, line 109: @dansan566 this must be validated by using the new reward_type ideal and comparing tzkt output with prpc output from a cycle with some lost blocks and endorements

Actually by comparing PRPC and tzkt for all 3 reward types after tzkt and tzstats have been confirmed of working the same way.

src/rpc/rpc_reward_api.py
```diff
@@ -107,37 +127,54 @@ def get_rewards_for_cycle_map(self, cycle, expected_rewards=False):
             # necessary data to properly compute rewards
             raise e from e
 
-    def __get_number_of_baking_rights(self, cycle, level):
+    def __get_baking_rights(self, cycle, level):
```
jdsika, line 130: @utdrmac I would ask you to please have a look at this section here

src/tzkt/tzkt_reward_api.py
```diff
@@ -69,25 +69,42 @@ def get_rewards_for_cycle_map(self, cycle, expected_reward=False) -> RewardProvi
...
+                + split['missedEndorsementRewards'] \
+                + split['uncoveredEndorsementRewards'] \
+                + split['missedOwnBlockFees'] \
+                + split['missedExtraBlockFees'] \
+                + split['uncoveredOwnBlockFees'] \
+                + split['uncoveredExtraBlockFees']
+            if rewards_type.isActual():
+                total_reward_amount = total_rewards_and_fees - total_equivocation_losses
+            elif rewards_type.isIdeal():
+                total_reward_amount = total_rewards_and_fees + total_offline_losses
```
jdsika, line 107: looks fine

src/tzstats/tzstats_reward_provider_helper.py
```diff
@@ -64,17 +65,26 @@ def get_rewards_for_cycle(self, cycle, expected_rewards=False):
...
+                                         + float(resp[idx_income_lost_accusation_rewards])
+                                         + float(resp[idx_income_lost_revelation_fees])
+                                         + float(resp[idx_income_lost_revelation_rewards]))
+            # losses due to being offline or not having enough bond
+            total_offline_losses = (float(resp[idx_income_missed_baking_income])
+                                    + float(resp[idx_income_missed_endorsing_income]))
+            if rewards_type.isActual():
+                root["total_reward_amount"] = int(1e6 * (total_rewards_and_fees - total_equivocation_losses))
+            elif rewards_type.isIdeal():
+                root["total_reward_amount"] = int(1e6 * (total_rewards_and_fees + total_offline_losses))
```
jdsika, line 88: looks fine

src/tzstats/tzstats_reward_provider_helper.py
```diff
@@ -64,17 +65,26 @@ def get_rewards_for_cycle(self, cycle, expected_rewards=False):
             raise ApiProviderException('GET {} {}'.format(uri, resp.status_code))
 
         resp = resp.json()[0]
-        if expected_rewards:
+        if rewards_type.isEstimated():
             root["total_reward_amount"] = int(1e6 * float(resp[idx_income_expected_income]))
```
jdsika, line 69: @dansan566 we have to make a reminder to talk to Alexander Eichhorn about the wording here. After our discussion I am inclined to say that the expectation by the client is a smooth working bakery and the "expected" API call from tzstats is in fact an "estimated"

tests/integration/test_tzkt_reward_api.py
```diff
@@ -75,13 +76,13 @@ def test_get_rewards_for_cycle_map(self, address, cycle):
                 nw=default_network_config_map['MAINNET'],
                 baking_address=address,
                 node_url='https://rpc.tzkt.io/mainnet')
-            rpc_rewards = rpc_impl.get_rewards_for_cycle_map(cycle)
```
jdsika, line 78: @nicolasochem only two reward types are tested here?

src/pay/payment_producer.py
nicolasochem: Yes, you are right.

src/rpc/rpc_reward_api.py
nicolasochem: On master branch currently, "expected rewards" and "ideal rewards" are equivalent and used interchangeably.

jdsika: Are all occurrences of expected rewards removed?

dansan566:
@nicolasochem I did some further testing and improved the Airtable. I've included also p RPC and found some interesting issues which require further investigation, especially with reward type "ideal": 
https://airtable.com/invite/l?inviteId=invTVGzEyebvnyyoH&inviteToken=939d76513d634125eeef86615e4d626a388bcf72ac6c9e08151374575b5fd99e&utm_source=email

src/rpc/rpc_reward_api.py
nicolasochem: The function `get_number_of_baking_rights` doesn't exist anymore, and has been replaced by `get_baking_rights`.

On master branch, I only need to count baking rights. But on this PR, I need to match the list of baking rights with the actual bakes, so I can tell which bakes were successful and which ones were not.

On this line though, I just need to count the rights. Before, the count was done inside the function, now it is done outside of the function.

src/rpc/rpc_reward_api.py
nicolasochem: Yes, because this whole code block is for estimated reward calculation. Ideal/actual rewards use another code block in the `else` case below.

tests/integration/test_tzkt_reward_api.py
nicolasochem: Yes, I didn't add ideal reward tests to tzkt api.

src/pay/payment_consumer.py
nicolasochem: Yes, it is better to have a default value here. I added it in d45441d. Thank you

nicolasochem:
> @nicolasochem I think in tzstats the rewards of prio 1 and higher blocks are already included in `(float(resp[idx_income_baking_income]) + float(resp[idx_income_endorsing_income])`

You and @dansan566 are right. I misunderstood tzstats api. This is now fixed with commit ee5b729. Thank you both.

> @nicolasochem I did some further testing and improved the Airtable. I've included also p RPC and found some interesting issues which require further investigation, especially with reward type "ideal":

This is great, let me have a look at these and get back to you.

nicolasochem:
> @nicolasochem I did some further testing and improved the Airtable. I've included also p RPC and found some interesting issues which require further investigation, especially with reward type "ideal":

## Row 9 and 12

in florence, the RPC changed, and the endorsement query is now `endorsement_with_slot` instead of just `endorsement`. 

See this line:

https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/415/files#diff-25ae46ce48f187f3cf93d7c4a0d36b6c79e96ca0dbacf22818fa3fc1181e3afaR175

On this line, if you replace `endorsement_with_slot` with `endorsement`, you will find that the ideal calculations for cycle 268 and 208 are correct. If you don't, it's going to incorrectly assume that all endorsements were missed, and add a ton of rewards to the ideal calculation.

If we merged this earlier, we would have add to add a check for cycle and handle pre-florence and florence. Now, every pre-florence block has been paid by everyone, so there is nothing we need to do. TRD "ideal" payout option does not work with pre-florence cycles anymore. If you can, you should find better testcases in cycles after florence.

## Rows 4, 5, 6

You wrote

> I expected ideal tzstats < tzkt because of the fees for the stolen block. this is not the case.

The fee for the stolen block is calculated appropriately. In the PR description, I describe a discrepancy between tzstats and tzkt in ideal mode when the blocks are **stolen from you**. If you steal blocks from someone else, there is no discrepancy.

## Rows 7, 8

> Why is ideal tzstats higher than tzkt?

That's because you had a block miss. For the same reason than stated in the PR description, tzstats ideal rewards takes into account the fees that the priority baker 1 that took your spot earned.

I could also ignore these extra fees so we have no discrepancy between sources. Any thoughts?

dansan566:
Row 9, 12, 4, 5, 6 make sense for me now! Thanks for explaining. However, I still have some troubles understanding 7 and 8. 
You wrote in your PR description:
> 
>These are the fees that the baker would have earned, had they baked the blocks that they missed. We include them in the calculation because it makes sense to do so. tzstats does not expose these lost fees, so there is an ideal reward discrepancy between tzkt and tzstats backend when the baker fails to bake one or more blocks in the cycle, with tzstats backend **reporting ideal payout lower** than its true value. But this is an edge case so it makes sense to enable ideal payouts with tzstats backend anyway. We documented it.
>
Lets assume we have a block miss then the #tzkt > #tzstats because the fee within the block will be added to #tzkt. In my test case, it is  #tzkt < #tzstats.

jdsika:
`in florence, the RPC changed, and the endorsement query is now endorsement_with_slot instead of just endorsement.`


@nicolasochem I think it is absolutely paramount that the program does work correctly within the specification! Either old cycles quit with an error OR we fix it by case handling which is the option I prefer! Eventually I think the RPC should behave the same way correctly for all cycles after a protocol upgrade - I consider this to be a protocol bug?

src/rpc/rpc_reward_api.py
nicolasochem: Yes. I grepped for `expected` and I get a few matches from tzstats and tzkt api since some API calls are named "expected". Also the integration test for tzkt has "expected" string but it is referring to "expected results" compared to "actual results" so that's different from "expected rewards".

nicolasochem:
> Row 9, 12, 4, 5, 6 make sense for me now! Thanks for explaining. However, I still have some troubles understanding 7 and 8.
> You wrote in your PR description:
> 
> > These are the fees that the baker would have earned, had they baked the blocks that they missed. We include them in the calculation because it makes sense to do so. tzstats does not expose these lost fees, so there is an ideal reward discrepancy between tzkt and tzstats backend when the baker fails to bake one or more blocks in the cycle, with tzstats backend **reporting ideal payout lower** than its true value. But this is an edge case so it makes sense to enable ideal payouts with tzstats backend anyway. We documented it.
> 
> Lets assume we have a block miss then the #tzkt > #tzstats because the fee within the block will be added to #tzkt. In my test case, it is #tzkt < #tzstats.

In cycle 268, you had [one missed block](https://tzstats.com/BMdc3xkAy4MZXc3wm3edrQ5utRHAZs5aTXJvu1FmbQg3DczpFM3). But someone did not endorse [the previous block](https://tzstats.com/BMbVy2DBswnvm47KDLqRA754bX97o4LvybQtJbqcYLpzhd2poFX). So the missed rewards are 38.75 tez in reality.

tzstats api returns 40 as missed baking income and tzkt returns 38.75.

Tzkt counts missed baking fees, but because it also counts the missed block reward more accurately, it ended up lower than tzstats.

These are small bugs in tzstats api. We can reach out to the dev team for them to fix them. But that should not prevent us from merging this PR.

nicolasochem:
> `in florence, the RPC changed, and the endorsement query is now endorsement_with_slot instead of just endorsement.`
> 
> @nicolasochem I think it is absolutely paramount that the program does work correctly within the specification! Either old cycles quit with an error OR we fix it by case handling which is the option I prefer!

Should TRD handle very old rewards from Athens then?

In general, we should handle protocol transitions seamlessly, but when the transitions are behind us, let's just remove the code. The old code is not helping anyone, but it may have bugs or just confuse people.

"old cycles quit with an error": if nothing changed in the calculation, I see no reason to do that. But if something has changed, I agree that it's better to terminate with an error than to give faulty results. I could do something like this:

https://github.com/nicolasochem/tezos-reward-distributor/compare/ideal_vs_expected...nicolasochem:ideal_vs_expected_pre_florence_fail

Please advise (I haven't tested the above code yet)

> Eventually I think the RPC should behave the same way correctly for all cycles after a protocol upgrade - I consider this to be a protocol bug?

No, this is a RPC change, which happens from time to time when the protocol upgrades.

jdsika:
> Should TRD handle very old rewards from Athens then?
> 
> In general, we should handle protocol transitions seamlessly, but when the transitions are behind us, let's just remove the code. The old code is not helping anyone, but it may have bugs or just confuse people.

ok, maybe we should add a note somewhere... maybe in the help section for RPC? "Important:..."

> "old cycles quit with an error": if nothing changed in the calculation, I see no reason to do that. But if something has changed, I agree that it's better to terminate with an error than to give faulty results. I could do something like this:
> 
> [nicolasochem/tezos-reward-distributor@ideal_vs_expected...nicolasochem:ideal_vs_expected_pre_florence_fail](https://github.com/nicolasochem/tezos-reward-distributor/compare/ideal_vs_expected...nicolasochem:ideal_vs_expected_pre_florence_fail)
> 
> Please advise (I haven't tested the above code yet)
> 
> > Eventually I think the RPC should behave the same way correctly for all cycles after a protocol upgrade - I consider this to 

hm, yes this might be a good test as it is clear that the number of endorsements should be retrievable and in this case it catches the protocol transition indirectly. Works for me!

dansan566:
@nicolasochem I also asked @akrykunenko (BakeNRolls) to help with testing. I was particular interested in cycle 374. It looks like that tzkt **pays out** accusation rewards within actual and ideal! RPC still needs to be tested.

nicolasochem:
> hm, yes this might be a good test as it is clear that the number of endorsements should be retrievable and in this case it catches the protocol transition indirectly. Works for me!

I just pushed it. This is what happens now when you try to payout stakenow cycle 268 with rpc method:

```
2021-07-28 09:49:35,774 - producer  - INFO - Checking for pending payments : checking 268 <= 385 - (5 + 1) - 0                                                                                                                                                
2021-07-28 09:49:35,774 - producer  - INFO - Payment cycle is 268                                                                                                                                                                                             
2021-07-28 09:49:35,775 - producer  - INFO - Using ideal rewards for payouts calculations                                                                                                                                                                     
2021-07-28 09:49:54,819 - producer  - WARNING - Found missed baking slot {'level': 1101172, 'delegate': 'tz1g8vkmcde6sWKaG2NN9WKzCkDM6Rziq194', 'priority': 0, 'estimated_time': '2020-08-26T12:17:06Z'}, adding 40000000 mutez reward anyway.                
2021-07-28 09:50:00,569 - producer  - ERROR - Can not parse endorsements from RPC. Aborting.                                                                                                                                                                  
2021-07-28 09:50:00,570 - producer  - INFO - TRD can not process rewards for protocols older than Florence.                                                                                                                                                   
2021-07-28 09:50:00,570 - producer  - ERROR - [try_to_pay] Generic exception Can not parse endorsements from RPC.                                                                                                                                             
2021-07-28 09:50:00,572 - producer  - ERROR - Unknown error in payment producer loop: Can not parse endorsements from RPC., will try again.                                                                                                                   
```

nicolasochem:
> @nicolasochem I also asked @akrykunenko (BakeNRolls) to help with testing. I was particularly interested in [cycle 374](https://tzkt.io/tz1NortRftucvAkD1J58L32EhSVrQEWJCEnB/rewards). It looks like that tzkt **pays out** accusation rewards within actual and ideal! RPC still needs to be tested but I expect a similar result as with tzkt. It is documented in the table as well.

Thanks @akrykunenko for helping with testing.

This issue is present in master as well:

https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/tzkt/tzkt_reward_api.py#L81-L82

It was done like this since the tzkt support was added.

We should not fix this within this PR. I think it should be configurable as said above.

jdsika:
@vkresch we had several strange failures for the validation_baker_address test. Is it using an external RPC? What can be the cause? TImeouts?

jdsika:
> This issue is present in master as well:
> 
> https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/tzkt/tzkt_reward_api.py#L81-L82
> 
> It was done like this since the tzkt support was added.
> 
> We should not fix this within this PR. I think it should be configurable as said above.

@nicolasochem do I understand that only tzkt has this behavior? In this case I would really prefer to modify the tzkt method so that it does not payout those accusation rewards in order for all methods to behave the same way and THEN make an issue to make this configurable in the future. I would say that this is my last note for this PR before I want to merge it. @dansan566 takes care about the descrepancies between tzkt and tzstats in a conversation with their devs. We will leave a note in the release notes about our observation.

nicolasochem:
@jdsika yes, it's only tzkt. I did what you asked in another PR #470

jdsika:
@nicolasochem I merged #470 and resolved the merge conflict of this branch. Please verify the comment I made for "ideal" and that I added the slashing rewards correctly in ideal.

dansan566:
looks like tzkt ideal is still broken!

jdsika:
@dansan566 yes, I understood from Nicolas that it should be part of ideal with every backend. But if you tell me it is now just part of tzkt I will remove it there again.

nicolasochem:
> @dansan566 yes, I understood from Nicolas that it should be part of ideal with every backend. But if you tell me it is now just part of tzkt I will remove it there again.

It was just part of tzkt. I pushed a fix so now accusation rewards are not part of any calculation for any reward type and any data source.

dansan566:
I did some final tests. All look good from my side. PR can be merged IMO. 

Now, I also know why tzstats and tzkt numbers can differ:

When an endorsement is included in a block N, the baker actually validates block N-1, as per Tezos docs. 
**tzkt:**
number of endorsements (slots) included **in** the current block.
**tzstats**
tzstats shows the number of validations **of** the current block. The head block always contains 0 endorsements and gets updated when they get confirmed next block.

Efforts: 7h

jdsika (approved):

Issue #417, opened 2021-04-15
Webhook Plugin incorrect types in payout.


priyeshs11:
**Describe the bug**
The example JSON from the plugin docs show all the `stakingBalance`, `amount` and the `feeAmount` to be **ints**, however, the plugin sends these amounts as floats in some cases. Here is an example of that from my webhook plugin logs:

```
could not decode tezos rewards body: json: cannot unmarshal number 8014999376.0 
into Go struct field payout.payouts.stakingBalance of type int64
```

**To Reproduce**
1. Run TRD with webhook plugin
2. Inspect the json sent by TRD

**Expected behavior**
`stakingBalance`, `amount` and the `feeAmount` should be **ints**

```
{
  "timestamp": 1604982374,
  "token": "Xynl6svphysd3BhjLP6IS",
  "subject": "Payouts of cycle 22 completed",
  "message": "Much longer message example containing more information",
  "payouts": [
    {
      "address": "tz1LrHNbbCLgNJZsEsTUYFvWz2THgJC8fHyX",
      "paymentAddress": "tz1LrHNbbCLgNJZsEsTUYFvWz2THgJC8fHyX",
      "addressType": "D",
      "cycle": 415,
      "stakingBalance": 65116664916,
      "ratio": 0.16080255,
      "feeRatio": 0.01398283,
      "amount": 207756875,
      "feeAmount": 18065815,
      "feeRate": 0.08,
      "payable": true,
      "skipped": false,
      "opHash": "oo4Gikxyj8cMqM8xzgWxqnsxXoGwfhZqHDrLqpA6ENmMVoYUnVd",
      "neededActivation": false,
      "paymentStatus": "DONE"
    },
    {...}
  ]
}
```

PR #418, opened 2021-04-16
Webhook type casting

vkresch:
---
name: Webhook type casting
about: Fix found user bug.
labels: 

---
Resolves https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/417

* **Analysis**: No type casting is performed hence creating type issues for users.

* **Solution**: Perform explicit type casting of the webhook payload.
* **Implementation**: Added a type cast function.

* **Performed tests**: Added regression test with ints, floats and strings.

* **Documentation**: Added comments to the regression test.

* **Check list**:

**Work effort**: 2h

### DIFF
--- a/tests/regression/test_webhook_send_admin_notification.py
+++ b/tests/regression/test_webhook_send_admin_notification.py
@@ -1,0 +1,75 @@
+import copy
+from src.Constants import PaymentStatus
+from src.plugins.webhook import WebhookPlugin
+from src.model.reward_log import RewardLog
+
+
+# Default rewardlog
+rewardlog = RewardLog(
+    "tz1LrHNbbCLgNJZsEsTUYFvWz2THgJC8fHyX",
+    "D",
+    staking_balance=65116664916,
+    current_balance=100000000000,
+)
+rewardlog.cycle = 415
+rewardlog.ratio = 0.16080255
+rewardlog.service_fee_ratio = 0.01398283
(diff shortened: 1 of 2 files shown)


tests/regression/test_webhook_send_admin_notification.py
```diff
@@ -0,0 +1,75 @@
...
+rewardlog2.current_balance = "1000000000"
+rewardlog3.amount = "12345678900"
+rewardlog3.service_fee_amount = "5555555"
+rewardlog3.ratio = "0.16080255"
+rewardlog3.service_fee_ratio = "0.01398283"
+rewardlog3.service_fee_rate = "0.08"
+
+cfg = {
+    "webhook": {
+        "endpoint": "https://testnet-tezos.giganode.io:443",
```
utdrmac, line 45: This is not correct. Tezos RPC's do not support webhook. You need to change this to the original testing server.

vkresch: done

utdrmac: will that actually work? mydomain.com/webhook.php ?

vkresch: It's only a dummy configuration which does no actual requests. 
The test only tests if the payload was generated correctly with the correct types.

utdrmac (approved):

PR #419, opened 2021-04-20
print traceback when payment producer crashes

nicolasochem:
I find it difficult to develop on payment producer code without the
traceback. Also, if it crashes in production, it will be easier to debug
the problem if the user can provide the full traceback.


habanoz (commented):
It is set to be "exc_info=True". It should be providing what "traceback.print_exc()" provides. 

Please see app.log file for exception details. Also a verbose log file is provided for finer details.

nicolasochem:
thanks for the info

PR #420, opened 2021-04-20
add ability to pay out frozen rewards with RPC

nicolasochem:
RPC has a helper to expose frozen rewards by cycle. This makes it
possible to pay out actual rewards before they unfreeze.

I am not sure if there is a specific reason this is not implemented? Is
it just that no one got to it, or is there a specific concern with these
values from RPC?


jdsika:
@nicolasochem we are currently a "bit behind" which will get better in two weeks (other work priorities...). Thank you for the contributions!

nicolasochem:
hi @jdsika can anyone look at this?

jdsika:
@amzid was the documentation adjusted to reflect the new capabilities of the RPC backend?

nicolasochem:
@jdsika it looks like all the integration tests run with [unfrozen rewards](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/tests/utils.py#L22) and there are no unit tests yet for `rpc_reward_api.py`.

Adding all of this is out of scope of the PR. I am finished.

Issue #410, opened 2021-04-07
Validate Baking Address


habanoz:
**Is your feature request related to a problem? Please describe.**
TRD validates payment addresses by checking if signer has the private key. However, there is no such validation for baking address.

**Describe the solution you'd like**
TRD should check if the baking address is a valid address at application launch and provide a clear description of what is not correct.

**Describe alternatives you've considered**
Validation logic may contain the following checks:
- Is the address present at all?
- Is the address enabled for delegation?

**Why does TRD need to validate baking address?**
I believe if the baking address is not correct, it should fail during launch. Trying to understand why an RPC request is failing would be harder.

utdrmac:
Why does TRD need to validate baking address?

habanoz:
The discussion can be found here: https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/405#issuecomment-815076473

I believe if the baking address is not correct, it should fail during launch. 

Trying to understand why an RPC request is failing would be harder.

utdrmac:
```
http://127.0.0.1:9731/chains/main/blocks/head/context/contracts/tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx/manager_key
<<<<1: 200 OK
{ "manager": "tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx",
  "key": "edpkuBknW28nW72KG6RoHtYW7p12T6GKc7nAbwYX5m8Wd9sDVC9yav" }
```

> Here, the key field contains a public key, which means a revelation operation has already been published.

https://www.ocamlpro.com/2018/11/15/an-introduction-to-tezos-rpcs-a-basic-wallet/

Issue #11, opened 2018-12-09
Unrevealed addresses need to be revealed


habanoz:
**Is your feature request related to a problem? Please describe.**
Some users are not that before using an address for transactions it needs to be revealed. Users should be made aware.

**Describe the solution you'd like**
The application may check and instruct users to run reveal operation. 

**Describe alternatives you've considered**
The application may transparently run reveal command, but it is better if users do it with awareness.

habanoz:
tzscan api is used for revelation check. An rpc based solution could not be found. 

Closing but suggestions are welcome.

vkresch:
I will open a PR for this like discussed in #422.

PR #422, opened 2021-04-20
Validate baking addresses

vkresch:
---
name: Validate baking addresses
labels: 

---

Resolves https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/410

* **Analysis**: TRD validates payment addresses by checking if signer has the private key. However, there is no such validation for baking address.

* **Solution**: TRD should check if the baking address is a valid address at application launch and provide a clear description of what is not correct.

* **Implementation**: Check if the baker address has revealed its public key (https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/410#issuecomment-816664409) and if it's delegatable for all APIs (rpc, tzkt and tzstats).

* **Performed tests**: Added regression tests for baker validation.

* **Documentation**: No Change.

* **Check list**:

**Work effort**: 8h


src/cli/client_manager.py
```diff
@@ -119,6 +119,48 @@ def check_pkh_known_by_signer(self, key_name, timeout=None):
...
+                                        timeout=timeout)
+        except Exception as e:
+            raise ClientException(f'Exception: {e}\n{baker_address_exception}')
+
+        if response.status_code != HTTPStatus.OK:
+            logger.error(f'{response.text}\n{baker_address_exception}')
+            return False
+
+        response_json = response.json()
+        if response_json is None or "delegate" not in response_json or response_json["delegate"] != address:
```
vkresch, line 159: Please correct me if it's a bad way to detect if an address is delegatable.

src/config/yaml_baking_conf_parser.py
```diff
@@ -156,8 +154,14 @@ def validate_baking_address(self, conf_obj):
 
         # key_name must has a length of 36 and starts with tz or KT, an alias is not expected
         if len(baking_address) == PKH_LENGHT:
-            if not baking_address.startswith("tz"):
-                raise ConfigurationException("Baking address must be a valid tz address")
+            if not baking_address.startswith("tz") and not baking_address.startswith("KT"):
```
vkresch, line 157: I added the KT address which can also pass because of the comment mentioned above. Is this fine?

src/cli/client_manager.py
habanoz: What if the endpoint returns 500? Is it correct to assume false for anything but 200?

RpcBlockApiImpl has another revelation operation. We better remove one of them. Also please correct me if I am wrong, the payment address also needs to be revealed but currently it is not checked. @amzid please comment.

src/config/yaml_baking_conf_parser.py
habanoz: I am not sure about this. Is a KT address delagatable?

habanoz (commented):
Please see my comments

src/config/yaml_baking_conf_parser.py
vkresch: Found a [resource](https://crypto.bi/tz-kt/): Originated addresses (TK) cannot act as bakers. Only manager addresses (tz) can be registered to bake.

src/cli/client_manager.py
vkresch: @habanoz thanks for mentioning the `RpcBlockApiImpl` Im still familiarizing myself with the codebase. I moved the validation from the client_manager into the yaml_conf_parser without any status code checking.

As I understand it correctly a reveal operation is not mandatory when you want to send tezos to this payment address. Thus a reveal is not a validator for a payment address, or?

habanoz (approved):

src/cli/client_manager.py
habanoz: The payee does not need to be revealed. However, the payer may need to be revealed. However, I am not sure. 

https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/11

I found a past issue that mentions revelation need.

src/config/yaml_baking_conf_parser.py
habanoz: This is a case that supports Robert Cecil Martin who claims "comments lie!" ;)

src/cli/client_manager.py
vkresch: @habanoz ok I will open a seperate PR for the payment address revelation check

PR #428, opened 2021-05-07
set signer request type to json

nicolasochem:
I am using [wrapper
code](https://github.com/midl-dev/tezos-remote-signer-os/blob/master/tezos-remote-signer/templates/usr/lib/python3/tezos-signer-wrapper/signerWrapper.py)
around my tezos-signer to expose health endpoints. It is requesting
application-type: json header. I am adding it here.

I tried it with a regular (non-wrapped) signer as well, this didn't
break it.

This is a tiny change which helps me and does not matter to anyone else.
Besides, when signing remotely, the tezos client adds this header in its
request to the signer; this change just makes TRD mimic this behavior.

### DIFF
--- a/src/cli/client_manager.py
+++ b/src/cli/client_manager.py
@@ -13,7 +13,6 @@ logger = main_logger
 COMM_BOOTSTRAP = "{}/monitor/bootstrapped"
 MAX_NB_TRIES = 3
 
-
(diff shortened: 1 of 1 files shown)


jdsika (approved):
Looks fine for me, @amzid any objections?

PR #433, opened 2021-05-11
run_mode ONETIME (3) should not run forever

nicolasochem:
I am integrating trd in
[tezos-on-gke](https://github.com/midl-dev/tezos-on-gke/) and bringing
it at feature parity with Backerei software.

In this operation model, we run the payout engine as a one-shot cronjob:
it will run on a cadence, and issue payouts (if any), else exit
immediately.

This mode is identical to backerei's default behavior. backerei
--continuous is equivalent to TRD mode 1 (run forever).

TRD does not have this capability currently: run_mode 3 (onetime) will
issue outstanding payout and exit, but if there are no outstanding
payouts, it will wait for next cycle.

This is odd behavior: when running a program, I should know whether it
will be long running, or short running. I believe run_mode 3 is of
limited use. People are most likely interrupting the process when they
find there is nothing outstanding.

Since you are collecting statistics, maybe you have more insight? If
this mode is useful, I am happy to introduce run_mode 5: pay oustanding
amounts for currenty cycle and exit, exit immediately if no outstanding
amounts.

### DIFF
--- a/src/pay/payment_producer.py
+++ b/src/pay/payment_producer.py
@@ -218,7 +218,7 @@ class PaymentProducer(threading.Thread, PaymentProducerABC):
                     logger.info("No pending payments for cycle {}, current cycle is {}".format(pymnt_cycle, current_cycle))
 
                     # pending payments done. Do not wait any more.
-                    if self.run_mode in [ RunMode.PENDING, RunMode.ONETIME ]:
+                    if self.run_mode in [RunMode.PENDING, RunMode.ONETIME]:
                         logger.info(f"{self.run_mode} satisfied. Terminating ...")
                         self.exit()
                         break
--- a/src/pay/payment_producer.py
+++ b/src/pay/payment_producer.py
@@ -218,8 +218,8 @@ class PaymentProducer(threading.Thread, PaymentProducerABC):
                     logger.info("No pending payments for cycle {}, current cycle is {}".format(pymnt_cycle, current_cycle))
 
                     # pending payments done. Do not wait any more.
-                    if self.run_mode == RunMode.PENDING:
-                        logger.info("Run mode PENDING satisfied. Terminating ...")
+                    if self.run_mode in [ RunMode.PENDING, RunMode.ONETIME ]:
+                        logger.info(f"{self.run_mode} satisfied. Terminating ...")
                         self.exit()
                         break
 


jdsika:
I think the idea is that someone starts the program before the cycle is finished, then the program is instructed to pay "1 cycle" and if the new cycle comes the program recognized it, pays and then does the exit. The "one time" wording is therefore chosen with regard to "one time payment" and not "one time check if something needs to be done"

We may have to think about the wording here. What is your opinion @habanoz ?

habanoz:
@nicolasochem 

Let me summarize all running modes.

The first option is to run TRD as a service or in interactive mode. 

In service operation mode, TRD is running uninterrupted in the background. The user does not need to worry about the upcoming cycles. 

In interactive operating mode, however, the user has full control and responsibility. Run modes only apply in interactive mode. 

Run Mode 1 (FOREVER) : This run mode is more suitable for service operation mode. It runs for outstanding cycles and any future cycles. 

Run Mode 2 (Pending) : TRD makes payments for any outstanding cycles and exits. 

Run Mode 3 (One Time) : TRD makes one payment and exits. It will wait if necessary. 

I believe your case is more suitable for run mode 2, pending. It will pay for any outstanding cycles and exit. Considering you use cronjobs, there should not be more than 1 pending cycle. If there is more than 1 cycle, then it should indicate an outage and even in that case, all pending cycles need to be paid. 

I am also interested in seeing how cronjobs fit the payout domain, considering that cycle endings are not periodic. Cycle ends are a function of the availability of baking nodes thus cannot be foreseen. 

> This is odd behavior: when running a program, I should know whether it will be long running, or short running. I believe run_mode 3 is of limited use. People are most likely interrupting the process when they find there is nothing outstanding.

In interactive mode, the user should know if a cycle is inbound or not. Considering the existence of the run mode pending,  I cannot think of a case in which a user starting at the one-time mode and then interrupting. Please explain your view.

Also, your PR makes ONE_TIME and PENDING modes similar thus removing the difference between the two modes. I cannot think a user would allow a cycle behind unpaid. 

I also think the one-time mode is of limited use, however, it is still an option one might prefer.

nicolasochem:
Thanks Husseyin, I didn't realize this was the only difference between 2 and 3. it looks like I should be using run mode 2, so I will.

You can have unsupservised operations with run mode 2 + cronjob just like run mode 1 as long as the program exits silently when there is nothing pending.

I have been running kubernetes payout cronjobs every few hours. I like this model. An advantage is that I can capture any abnormal exit (via exit code) and send it to an alerter without relying on TRD's native alerting system. And also it adds some randomness to the payout timing within a cycle. And you don't have a process running all the time. Other than that, there is no difference really.

In this context, a mode which "runs just for one cycle (if any) and exits" makes sense. In case of outage, you would need to run several times to catch up, but if your cronjob runs much more frequently than tezos cycles, it will catch up.

This is the behavior I was expecting from run mode 3 after reading the help and doc. I got confused. Based on your explanations, run mode 3 is suited for an user who (1) logins every 3 days to a window multiplexer (screen, tmux), (2) looks at the result of past payout and (3) when satisfied, issues the next payout.

But who does that? I would expect the operator to just let things run unless there is a problem. Could you check statistics for run mode 3 and maybe remove it?

habanoz:
@nicolasochem I checked the collected statistics, run mode details are not known. Only service mode usage is tracked. 

You can also access the statistics. 

https://trdstats.s3-us-west-2.amazonaws.com/index.html

Issue #429, opened 2021-05-08
"retry failed payments and exit" does not honor "initial cycle"


nicolasochem:
**Describe the bug**
"retry failed payments and exit" does not honor "initial cycle"

**To Reproduce**
1. have payments fail for one cycle N (for example due to lack of funds)
2. set "initial cycle" to N+1
3. run with `-M 4` "retry failed payments and exit"
4. observe that it is again attempting to pay for cycle N

**Expected behavior**
It should not attempt to pay failed payments for cycle N.

nicolasochem:
I see the same issue with run_mode 2. It tries to pay any failed past payment before proceeding to the current unpaid ones. I don't think it should do that.

PR #434, opened 2021-05-11
make default initial cycle take into account release_override

nicolasochem:
When no --initial_cycle and --release_override parameters are passed to
TRD,  assuming no previous payments have been
made, TRD will attempt to pay rewards for the cycle that most recently
unfroze.

If current cycle is 358, it will pay rewards for cycle 352.  This is normal
and expected behaviour.

The issue is that this behavior is not altered when --release_override
is used. If you want to pay rewards as soon as the cycle is
completed (--release_override -5) or as soon as the rights are
calculated (--release_override -11 and pay ideal rewards),
TRD STILL attemts to pay rewards for the cycle that most recently
unfroze when no --initial_cycle is given.

For example, if current cycle is 358:

* run TRD with `--release_override -5`, no prior payments (reports dir
  is empty) and no `--initial_cycle. It tries to pay rewards for cycle
  352, but the expected behavior is to attempt to pay cycle 357.
* run TRD with `--release_override -11`, no prior payments (reports dir
  is empty) and no `--initial_cycle. It tries to pay rewards for cycle
  352, but the expected behavior is to attempt to pay cycle 363.

This PR fixes this by taking into account release_override when
calculating the initial payment cycle when none is given.

Users explicitly passing a value to --initial_cycle are not impacted.
Users with previous succesful payments in their reports dir are not
affected. This only affects first-time users who set --release_override.

This is what I believe the user expects, but note that I am not
familiar with all the ways people run TRD, so reviewers are invited to
carefully consider the implications of this change based on their
interactions with users and their collected statistics.

In this PR I also edit some comments to make them clearer based on what
I discovered while reading code.

### DIFF
--- a/src/launch_common.py
+++ b/src/launch_common.py
@@ -84,7 +84,7 @@ def build_parser():
 def add_argument_cycle(parser):
     parser.add_argument("-C", "--initial_cycle",
                         help="First cycle to start payment. For last released rewards, set to 0. Non-positive values "
-                             "are interpreted as: current cycle - abs(initial_cycle) - (NB_FREEZE_CYCLE+1). "
+                             "are interpreted as: current cycle - abs(initial_cycle) - (NB_FREEZE_CYCLE+1) - release_override. "
                              "If not set application will continue from last payment made or last reward released.",
                         type=int)
 
--- a/src/pay/payment_producer.py
+++ b/src/pay/payment_producer.py
@@ -139,7 +139,7 @@ class PaymentProducer(threading.Thread, PaymentProducerABC):
             return
 
         # if non-positive initial_payment_cycle, set initial_payment_cycle to
-        # 'current cycle - abs(initial_cycle) - (NB_FREEZE_CYCLE+1)'
+        # 'current cycle - abs(initial_cycle) - (NB_FREEZE_CYCLE+1) - self.release_override'
         if self.initial_payment_cycle <= 0:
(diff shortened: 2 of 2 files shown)


jdsika:
The only issue I see is that currenty if some payments are skipped in a cycle, e.g. because the reward was lower then the fee to be paid then the successful payments and the unsuccessful payments (with can be argued if they are "unsuccessful" as they are skipped by intention) are in the same file within the "failed" folder. I actually do want those two to be split into to files. The successful payments in the "done" folder and the skipped/failed payments in the "failed" folder. This way your logic would recognize a successfully payed cycle even if some elements are still in the "failed" folder?

@amzid what do you think?

jdsika:
Tagging #429 here

amzid:
> The only issue I see is that currenty if some payments are skipped in a cycle, e.g. because the reward was lower then the fee to be paid then the successful payments and the unsuccessful payments (with can be argued if they are "unsuccessful" as they are skipped by intention) are in the same file within the "failed" folder. I actually do want those two to be split into to files. The successful payments in the "done" folder and the skipped/failed payments in the "failed" folder. This way your logic would recognize a successfully payed cycle even if some elements are still in the "failed" folder?
> 
> @amzid what do you think?

I also agree with the idea, but I would shift this to #429 or a new issue.

amzid (approved):

jdsika:
Ok, then I merge and we make another PR with the split that also resolves #429

Issue #435, opened 2021-05-12
Add skipped status + isolate failed/done


amzid:
* Mark skipped payment items as so instead of marking them as failed or done (+ add reason for skipping)
* Isolate done payments from skipped/failed ones

jdsika:
so the "retry failed" mode on that cycle should honor the existing successful payments, not override them, only try the ones from the "failed" folder and append the successful ones to the existing log?

jdsika:
we need to have a look at #388 as well to see if the status of injected but no positive confirmation is also handled correctly in the split of the two reports

habanoz:
> * Mark skipped payment items as so instead of marking them as failed or done (+ add reason for skipping)
> * Isolate done payments from skipped/failed ones

I believe, it is necessary to explain why these changes are necessary? What is the problem or shortcoming here?

amzid:
> we need to have a look at #388 as well to see if the status of injected but no positive confirmation is also handled correctly in the split of the two reports

Only FAILED payments will be under failed report. INJECTED and SKIPPED payments will be under done report to avoid confusion.



> > * Mark skipped payment items as so instead of marking them as failed or done (+ add reason for skipping)
> > * Isolate done payments from skipped/failed ones
> 
> I believe, it is necessary to explain why these changes are necessary What is the problem or shortcoming here?

The reason for the first change was explained in the PR. The reason for the second change is that many bakers found it misleading to have successful transactions under the failed report directory. So with this change there will be a split between successful and failed payment items.

PR #438, opened 2021-05-14
Resolve issue #435

amzid:
This PR resolves the issue #435. The following steps were performed:

* **Implementation**:  
   * Separate done and failed payouts 
   * Add skipped status
   * Improve tzstats implementation for retry failed
@habanoz would be great if you have a look if the new changes are well-defined according to the state machine

* **Performed tests**: dry run in mainnet.

* **Documentation**: comments in code

**Work effort**: 3 hrs.


jdsika:
btw, how can people override the decision that payments are skipped with "fee > payout amount" ?

src/Constants.py
```diff
@@ -58,6 +58,7 @@ class PaymentStatus(Enum):
     PAID = 1
     DONE = 2
     INJECTED = 3
+    SKIPPED = 4
```
habanoz, line 61: SKIP is a term used in calculations. I find this confusing, a different name may be used. That is being said, What is the distinction between skipped and fail?

src/pay/batch_payer.py
```diff
@@ -429,6 +429,7 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
 
             # if pymnt_amnt becomes 0, don't pay
             if pymnt_amnt == 0:
+                payment_item.paid = PaymentStatus.SKIPPED
```
habanoz, line 432: I would argue DONE is better suited for this case.

src/tzstats/tzstats_reward_provider_helper.py
```diff
@@ -177,8 +177,12 @@ def get_rewards_for_cycle(self, cycle, expected_rewards=False):
...
     def update_current_balances(self, reward_logs):
         """External helper for fetching current balance of addresses"""
-        for rl in reward_logs:
-            rl.current_balance = self.__fetch_current_balance([rl.address])[rl.address]
+        split_addresses = split(reward_logs, 50)
+        for list_address in split_addresses:
+            addresses = [x.address for x in list_address]
+            list_curr_balances = self.__fetch_current_balance(addresses)
+            for d in list_address:
+                d.current_balance = list_curr_balances[d.address]
```
habanoz, line 186: I am unable to understand the reason behind this change.

habanoz (changes requested):
@amzid Please see my comments. I will do another review according to your reply.

habanoz:
I have made a review, waiting for a reply to make another review. I will check state machine diagrams.

Please see my comment on the linked issue.
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/435#issuecomment-841720725

habanoz:
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/435#issue-889940928

I do not know if it is part of the request but, the reason for skipping is not added anywhere.

habanoz:
Review effort is 1H.

src/tzstats/tzstats_reward_provider_helper.py
amzid: This is to improve the implementation of tztats for retry_failed mode, mainly the way that current balances are updated is done in batches now instead of single queries

src/Constants.py
amzid: * PAID/DONE status is for payment items which are successfully applied and are (or are expected to be) part of the blockchain and can be queried using a block explorer. When a delegator has a payout with a status PAID that means the baker is able to provide a proof on the blockchain that the reward payment was done successfully.
* FAILED status is for payment items which failed at any stage of the payment pipeline (run_operation/forging/signing/pre_apply). The reasons for that depend on the individual case and bakers in general need support to solve that (support queries take time and effort).
* SKIPPED status is for payment items which are avoided with intention. In most cases due to the fact that costs of the transaction are higher than the reward amount. For many bakers, this transaction is not failed because it is avoided with intention and it is not DONE because the payout was not really performed by the baker, thus the idea of this new status.

src/pay/batch_payer.py
amzid: SKIPPED status is for payment items which are avoided with intention. In most cases due to the fact that costs of the transaction are higher than the reward amount. For many bakers, this transaction is not failed because it is avoided with intention and it is not DONE because the payout was not really performed by the baker, thus the idea of this new status.

vkresch:
@amzid please change the baking address of `tz1Z1tMai15JWUWeN2PKL9faXXVPMuWamzJj` (TezosLand) to `tz1g8vkmcde6sWKaG2NN9WKzCkDM6Rziq194` (StakeNow) in tezos-reward-distributor/tests/unit/test_BakingYamlConfParser.py because TezosLand is not a valid baker anymore thus the CI is failing. Maybe we should patch the baker validation function.

jdsika:
@vkresch sorry, could you fix it as he seems not to be responding?

amzid:
@vkresch @jdsika I have just changed the test baking address! Thanks for the review!

jdsika (dismissed):

habanoz:
@amzid 

> For many bakers, this transaction is not failed because it is avoided with intention and it is not DONE because the payout was not really performed by the baker

DONE is used for zero amounts. It is not necessary to make any payments but as a task, it is completed. DONE is also used in the DRY mode for the same interpretation. If payout was performed final status would be PAID. 

> We could change the new of SKIPPED to AVOIDED or SKIPPED_DUE_TO_HIGH_FEES if you think that the name is misleading?

I am OK with new names. You may choose one of them.

src/pay/batch_payer.py
habanoz: I think the new status is better after seeing your comments.

habanoz, line 432: Since the payment amount is 0, this I believe should be DONE.

habanoz:
>  if you have a look if the new changes are well-defined according to the state machine

For state changes I have created https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/443

habanoz:
Total effort 1H +0.5 H = 1.5 H

amzid:
Thank you for the feedback and explanations @habanoz. I also agree with the outcome. I will revert the done status for zero amounts and rename the new status.

habanoz (approved):

PR #440, opened 2021-05-20
fix #429 do not attempt to retry failed payments before INITIAL_CYCLE…

### LINKED ISSUE #429: "retry failed payments and exit" does not honor "initial cycle"

nicolasochem:
… given

I added test coverage for this new behavior.


jdsika:
My take on it with two options:
1) We either allow the runmode to be only usable with a specified cycle. The initial_cycle argument is then irrelevant and adding a cycle is mandatory. The user can not by accident trigger multiple payouts.
2) Multiple payouts are allowed as it is. Then you can either specify a specific cycle or without this argument all failed cycles will be retried. In this case the initial_cycle argument should be considered here as this PR says.

I am pro option 2) and would ask @nicolasochem to add a warning at the beginning that says something like "attention you are not specifiying a cycle which results in the retry of all failed payments - use this option with care".

nicolasochem:
@jdsika if you pass `-C` you are specifying the initial cycle, but option 4 will ignore this in master today. My PR makes it take this argument into account.

In Master, option 4 looks for any failed payment and retries it. In my PR, option 4 looks from any failed payment for a cycle >= initial cycle if given.

You can not "retry failed payment for one cycle and exit" today. I can add a warning, but then it would be printed all the time. Or are you saying we need an option 5 "retry failed payment for one cycle and exit"? With this option 5, you could use -C to specify which cycle you want to retry.

habanoz:
@nicolasochem Please check the test failures. We cannot proceed when there are test failures.

habanoz:
@nicolasochem "Initial payment cycle" is a core parameter and care is needed here. 

Initial payment cycle has multiple possible value ranges: negative, positive and zero. Also initial payment cycle is implicitly assigned if there is at least one payment record under the success directory (last_success + 1).

Could you please describe the proposed behavior of the retry producer script under those circumstances?

nicolasochem:
@habanoz thank you for the reminder.

> Also initial payment cycle is implicitly assigned if there is at least one payment record under the success directory (last_success + 1).

I actually don't think it's right. I have opened another PR to discuss #447 . We can stash this PR until #447 is resolved.

jdsika:
> You can not "retry failed payment for one cycle and exit" today. I can add a warning, but then it would be printed all the time. Or are you saying we need an option 5 "retry failed payment for one cycle and exit"? With this option 5, you could use -C to specify which cycle you want to retry.

@nicolasochem  I think giving a specific initial_cycle value and having 0 (the latest released reward) as the default value is ok. I think we will better not introduce another run mode

nicolasochem:
> > You can not "retry failed payment for one cycle and exit" today. I can add a warning, but then it would be printed all the time. Or are you saying we need an option 5 "retry failed payment for one cycle and exit"? With this option 5, you could use -C to specify which cycle you want to retry.
> 
> @nicolasochem Can you please double check this in combination with #458 which I want to merge first? I think the case of -1 needs to be handled here!

It is already handled. I am adding a check that the failed cycle is >= initial_cycle, so if initial_cycle is -1 it will always work (just like 0 used to work).

jdsika:
But -1 means "starting from last released one". My problem is that I THOUGHT that the initial_payment_cycle (which is initial_cycle) impacts "pymnt_cycle" in the run() function which is executed by starting the thread. You are giving the parameter to the function that is only read from the arguments CLI. I would have thought that the expected behaviour by not setting -C (so default -1) is to to the retry failed from the calculated value "pymnt_cycle"? 
What am I missing here?

nicolasochem:
There are 2 things:

1. if you pass --intial_cycle X, it won't do anything before cycle X
2. if you don't pass --initial_cycle X, it won't do anything before recently released cycle

In this PR I am ensuring 1.

I am not ensuring 2.

For 2 I have a better idea, but that's for another PR: we should make `--initial_cycle` mandatory when retrying. Then it will only retry the cycle given and exit. You can't "retry every cycle" because it's dangerous.

This PR is a progress already because we are ensuring 1. 2 will come later.

jdsika (approved):

PR #447, opened 2021-06-19
do not catch up on all late payments by default

nicolasochem:
The current behavior of TRD is to pay out every late payment at boot
time. So if my TRD did not run for a while, and I start it again, it
will pay everything it missed.

It may feel right, but it is not. Bakers almost certainly do not want to
run several payment cycles at once without interaction.

In regular TRD usage, you are not late. You want to pay the outstanding
payout for the cycle that has just been released.

If you happen to be late, you are probably not going to just restart the
software and call it a day.

Instead, you will be doing every payout manually, verifying after each
one and communicate with your delegates as you go, until the
situation is resolved.

So the current behavior on master is not useful. Most probably no one
wants this. But there is a chance that it happens by accident. Say, if
for some reason you lost most of your payout csv files until some date
in the past, and you do not pass the `--initial_cycle` parameter, you
will drain your payout account.

Therefore I propose to:

* not pay every payment since last success by default
* instead, assume `initial_payout_cycle=0` by default, which means that
  TRD will pay the recently released cycle only.
* you can still override this and pay out several unpaid cycles at once
  by passing a negative value to `initial_payout_cycle`, or an old
  cycle number.

I also removed some dead code that I found while doing this. There were
3 functions named `get_latest_report_file`, only one of which was used.

**Work effort**: Compensated otherwise


dansan566:
@nicolasochem I do like the PR very much. However, I always use `-C CycleNumberWhichIwanttoPayOut`, for example `-C 371`. Is this still supported? Also when using `-M 4`, the `-C` command should always be mandatory (or only use the latest failed cycle) IMO. If you have some failed files in pymts/bakeraddress/payments/failed, TRD does not care about `-C` which is not the behavior I would expect.

jdsika:
> 
> 
> @nicolasochem I do like the PR very much. However, I always use `-C CycleNumberWhichIwanttoPayOut`, for example `-C 371`. Is this still supported? Also when using `-M 4`, the `-C` command should always be mandatory (or only use the latest failed cycle) IMO. If you have some failed files in pymts/bakeraddress/payments/failed, TRD does not care about `-C` which is not the behavior I would expect.

The fun part: According to the help the way you use it by typing the "I want to pay out cycle 5678" is not working this way.

```
First cycle to start payment. For last released
                        rewards, set to 0. Non-positive values are interpreted
                        as: current cycle - abs(initial_cycle) -
                        (NB_FREEZE_CYCLE+1) - release_override. If not set
                        application will continue from last payment made or
                        last reward released.

```

jdsika:
@nicolasochem what do you think about a note from TRD that says:

"I am detecting that you are paying out cycle 360 but I see that you only have done payments reports until 350" 
or
"350 payment reports from 360 are present"

jdsika:
`def get_latest_report_file(payments_root)` this was implemented three times? ...

src/launch_common.py
```diff
@@ -83,18 +83,21 @@ def build_parser():
...
     parser.add_argument("-C", "--initial_cycle",
-                        help="First cycle to start payment. For last released rewards, set to 0. Non-positive values "
-                             "are interpreted as: current cycle - abs(initial_cycle) - (NB_FREEZE_CYCLE+1) - release_override. "
-                             "If not set application will continue from last payment made or last reward released.",
-                        type=int)
+                        help="First cycle to start payment(s) from depending on run mode -M. "
+                             "The default value is 0, which will start at the most recently released reward. "
+                             "Positive values sets desired cycle directly: initial_cycle <= current cycle - (NB_FREEZE_CYCLE+1) - release_override. "
+                             "Non-positive values count backwards: current cycle - abs(initial_cycle) - (NB_FREEZE_CYCLE+1) - release_override.",
+                        default=0, type=int)
```
jdsika, line 90: @nicolasochem I added the default value here as all other options have them here as well. I tried to adapt the help section of -C and the confusing part is that it does not work consistently in the ranges from negativ over 0 to positive values. What do you think?

jdsika: In addition I think we have to restrict the value range of release_override from [-11,-1] as it crosses with the initial_cycle defintion, or?

jdsika:
@nicolasochem I will merge after you have given your opinion to my questions

nicolasochem:
> `def get_latest_report_file(payments_root)` this was implemented three times? ...
> 
> Should we just leave the implementation in `src/util/dir_utils.py` for possible further useage as it is a "utils collection"?

In my opinion, if the code is not used anywhere, it should not be in the source tree.

> @nicolasochem I do like the PR very much. However, I always use -C CycleNumberWhichIwanttoPayOut, for example -C 371. Is this still supported?

Yes, this still works.

> Also when using -M 4, the -C command should always be mandatory (or only use the latest failed cycle) IMO. If you have some failed files in pymts/bakeraddress/payments/failed, TRD does not care about -C which is not the behavior I would expect.

I agree. PR #440 fixes this.

> what do you think about a note from TRD that says: "I am detecting that you are paying out cycle 360 but I see that you only have done payments reports until 350" or "350 payment reports from 360 are present"

I don't like it so much. If a baker missed 10 payouts, they will know. They don't need TRD to tell them.

> I added the default value here as all other options have them here as well. I tried to adapt the help section of -C and the confusing part is that it does not work consistently in the ranges from negativ over 0 to positive values. What do you think?

Sure. The effect is the same: initial_cycle is set to zero when not specified, so it will not "catch up" on past payments unless you explicitly set a negative initial_cycle or an absolute cycle number older than most recently released cycle.

src/launch_common.py
nicolasochem: @jdsika I see what you tried to do by fixing the help message but I think it fell short. I pushed another commit on top of yours. I think it makes the help message more helpful, please let me know what you think. Please merge if it's ok.

>In addition I think we have to restrict the value range of release_override from [-11,-1] as it crosses with the initial_cycle defintion, or?

Yes, it's probably out of scope for this PR but should be done.

jdsika: I just added it. Otherwise we will forget about it. I noted it in the help section and extended the value check after parsing.

jdsika (approved):

PR #458, opened 2021-07-21
fix CLI argument default values

jdsika:
I inserted a check for release_override to be in the range of [-11,-1] to take effect and have 0 as default

I removed the option for negative initial_cycle to set a default to -1 which results in "last released cycle" and [0,) to set a cycle by the user. Rationale: -1 is the cycle for the genisis block which is not relevant for payouts. The definition of -C was confusing and unnecessary

TODO:
[x] run.rst from the docs is still outdated
[x] configure.py needs updated parsing actions for correct ranges

Effort: 2h


jdsika:
@denver-s can you check this one, please? Current master is not working as I inserted the check for release_override incorrectly forgetting about the default value 0. 

I changed the defintion of initial_cycle to be better understandable and have less strange options.

src/pay/payment_producer.py
```diff
@@ -126,17 +126,20 @@ def run(self):
...
-            logger.debug("Payment cycle is set to {}".format(pymnt_cycle))
+        # if initial_payment_cycle has the default value of -1 resulting in the last released cycle
+        if self.initial_payment_cycle == -1:
+            pymnt_cycle = current_cycle - (self.nw_config['NB_FREEZE_CYCLE'] + 1) - self.release_override
+            if pymnt_cycle < 0:
+                logger.error("Payment cycle cannot be < 0 but configuration results to {}".format(pymnt_cycle))
+            else:
+                logger.debug("Payment cycle is set to last released cycle {}".format(pymnt_cycle))
+        else:
+            pymnt_cycle = self.initial_payment_cycle
```
denver-s, line 142: If I am not wrong, with this change `release_override` is taken into account only if `initial_payment_cycle = -1` (the default value).

In this case, if I want to start paying from (say) cycle 370, and I set the `release_override`, it won't be taken in consideration?

nicolasochem:
On master branch there are 3 possibilities for `--initial_cycle`

Positive: absolute cycle number to pay
Zero: pay recently released cycle (depends on release override parameter, but if release_override is zero, it would be the cycle before the current one)
Negative: pay cycle older than the recently released one. For example, assuming release_override of 0, `--initial_cycle -1` would pay 2 cycles before the current one.

You are suggesting to **remove** the ability to give a negative value.

I agree. I think this positive/negative/zero behavior is very confusing.

But you are not going far enough.

If negative values go away, then why have "zero" an option? Instead, we could just have "positive" and "none".

This is how the help would look like:

```
    parser.add_argument("-C", "--initial_cycle",
                        help="Cycle to start payment(s) from."
                             "If ommitted, will pay rewards that were most recently released."
                             "Cycle for which rewards were most recently released is calulated based on the formula:"
                             " current_cycle - (NB_FREEZE_CYCLE+1) - release_override"
                        type=int)
```

@habanoz what do you think? Is it OK to remove the ability to give negative values?

jdsika:
> 
> 
> On master branch there are 3 possibilities for `--initial_cycle`
> 
> Positive: absolute cycle number to pay
> Zero: pay recently released cycle (depends on release override parameter, but if release_override is zero, it would be the cycle before the current one)
> Negative: pay cycle older than the recently released one. For example, assuming release_override of 0, `--initial_cycle -1` would pay 2 cycles before the current one.
> 
> You are suggesting to **remove** the ability to give a negative value.
> 
> I agree. I think this positive/negative/zero behavior is very confusing.
> 
> But you are not going far enough.
> 
> If negative values go away, then why have "zero" an option? Instead, we could just have "positive" and "none".
> 
> This is how the help would look like:
> 
> ```
>     parser.add_argument("-C", "--initial_cycle",
>                         help="Cycle to start payment(s) from."
>                              "If ommitted, will pay rewards that were most recently released."
>                              "Cycle for which rewards were most recently released is calulated based on the formula:"
>                              " current_cycle - (NB_FREEZE_CYCLE+1) - release_override"
>                         type=int)
> ```
> 
> @habanoz what do you think? Is it OK to remove the ability to give negative values?

There is a cycle zero if you have a look at tzkt. Considering the fact that TRD can be used on test networks we have to make sure it works correctly from the start. Cycle -1 is according to tzkt assigned to the gensis block which is created out of thin air with no baking associated. I thought that the range [0,) is then correctly assigned and the value -1 can be safely used for the explizit wish to "start from last released cycle"

src/pay/payment_producer.py
jdsika: Look at line 179. I think it is taken correctly into account. If you want to explicitly mention a cycle to start with it should not be altered by another option anyway

nicolasochem:
You removed negative `--initial_cycle`. That's good. Now you can do even better and remove `--initial_cycle -1`

Instead, have no default value. When the user does not use `--initial_cycle` option, pay most recently released cycle.

It's in line with your PR, but even better. `--initial_cycle -1` is confusing.

jdsika:
The C++ guy in me does not Not Assign values to variables... It is an undefined state

src/pay/payment_producer.py
nicolasochem: Yes, I don't see a problem here.

nicolasochem (approved):
Fine, let's keep -1 as default value.

denver-s (approved):

jdsika:
@denver-s I gave you a standard 0.5h review time

Issue #444, opened 2021-06-07
KT addresses pay commission on the network, TZ addresses do not.


SemElVik:
what is the problem?
Why do my KT addresses pay for the commission on the network, but my tz addresses do not pay it?
what has changed? from cycle 352
how can i fix it
https://baking-bad.org/KT1XVS8SYD5A3TSHg22XjBvUWdhF3UFYg77Q
https://baking-bad.org/KT1TCcxKHVbn8CVoRVqqixCFSbrK8zjd3AP1

https://baking-bad.org/tz1e2WhFEHXjrxUzTPDZNXRK1xdw8jpgYc7q
https://baking-bad.org/tz1UcWDfjBHJRctBipVVzs2L77Pu7eH8xwYX (edited) 

my settings have not changed and now they are
---------
```reactivate_zeroed : True
delegator_pays_xfer_fee: False
delegator_pays_ra_fee : True
rules_map: {tz...........: TOF}
rewards_type: actual```

amzid:
This was originally changed in https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/7381a10d99a9ac1de18cf6d3a4f01d8927768263 to avoid too high fees for KT1 addresses, but apparently the fee is always substructed from the payment amount independently from the `delegator_pays_xfer_fee` setup. Can you double check this please @utdrmac?
![image](https://user-images.githubusercontent.com/33556352/124395957-4fe74c00-dd07-11eb-8c07-ecc7662731d7.png)

SemElVik:
Maybe I just need to increase the commission for the newsletter?  

0.000574 (tz) + 0.000873 (KT) =   0.001447 XTZ  
`fee.ini`  = 1447
it's a crutch but....

SemElVik:
will this be fixed?

jdsika:
Yes, Wednesday most likely

SemElVik:
[https://baking-bad.org/KT1U6Kdd42ZW3NxY2YnVV66SdjnbccRn1Zdh](url)

the commission has been reduced, but the problem has not been resolved

PR #461, opened 2021-07-21
Fix KT1 pay for transfer fee

denver-s:
---
name: Fix KT1 pay for transfer fee
about: Fixing a wrong calculation of transaction fee for KT1 delegators
labels: 

---
IMPORTANT NOTICE:

This PR resolves the issue #444. The following steps were performed:

* **Analysis**: It seems that batch_payer was not taking into account the possible `delegator_pays_xfer_fee` value.

* **Solution**: Make the transfer fee logic valid for all delegators kind.

* **Check list**:

- [x] I created a new issue if there is further work left to be done (if needed).

**Work effort**: 30 minutes.


src/pay/batch_payer.py
```diff
@@ -432,6 +434,8 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
                 payment_item.paid = PaymentStatus.DONE
                 logger.info("Payment to {:s} became 0 after deducting fees. Skipping.".format(payment_item.paymentaddress))
                 continue
+            else:
+                logger.info("Payment to {:s} became {:10.6f} after deducting fees.".format(payment_item.paymentaddress, pymnt_amnt / MUTEZ))
```
amzid, line 438: I think it is better to avoid this, this will be printed for every single transaction which is a lot, furthermore if no fees are deducted this might be misleading for the user ("after deducting fees").

src/pay/batch_payer.py
```diff
@@ -393,37 +393,39 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
...
                     continue
 
-                # Fees within safety limits; Subtract from payment amount
+                # Subtract burn fee from the payment amount
                 orig_pymnt_amnt = pymnt_amnt
-                pymnt_amnt = max(pymnt_amnt - total_fee, 0)  # ensure not less than 0
-                logger.info("Payment to {} script requires {:.0f} gas * {:.2f} mutez-per-gas + {:10.6f} default fee = {:10.6f} total fees; Payment reduced from {:10.6f} to {:10.6f}".format(
-                            payment_item.paymentaddress, gas_limit, MUTEZ_PER_GAS_UNIT, self.default_fee / MUTEZ, total_fee / MUTEZ, orig_pymnt_amnt / MUTEZ, pymnt_amnt / MUTEZ))
+                pymnt_amnt = max(pymnt_amnt - burn_fee, 0)  # ensure not less than 0
+                logger.info("Payment to {} script requires {:.0f} gas * {:.2f} mutez-per-gas + {:10.6f} burn fee; Payment reduced from {:10.6f} to {:10.6f}".format(
```
amzid, line 413: The printed info is not consistent with the code: only the burn fee is substructed in this case.

src/pay/batch_payer.py
```diff
@@ -393,37 +393,39 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
...
+                            payment_item.paymentaddress, gas_limit, MUTEZ_PER_GAS_UNIT, burn_fee / MUTEZ, orig_pymnt_amnt / MUTEZ, pymnt_amnt / MUTEZ))
```
amzid, line 415: There is one more check which is **specific to KT1 accounts**. Since fees can be really large even for small amounts, we would like to avoid payments with **total_fee > pymnt_amnt**. This was indirectly included in the previous logic, but needs to be implemented explictly now with a small if statement.

amzid (changes requested):

src/pay/batch_payer.py
denver-s: Can this be just a debug log?

amzid: I changed it to debug and added the safety check for fee amount. It would be great if you would double check/test it!

denver-s: Thanks! I'll do it ASAP.

src/pay/batch_payer.py
denver-s: Isn't it true that the same safety check would have been done by the following code (lines 436-438)?

```
# Subtract transaction's fee from the payment amount if delegator has to pay for it
if self.delegator_pays_xfer_fee:
    pymnt_amnt = max(pymnt_amnt - tx_fee, 0)  # ensure not less than 0
```

denver-s: The rational is that:

- KT1 address: subtract burn fee
- tz1 address: subtract reactivation fee

For both:
- subtract the transaction fee, and check if amount > 0

amzid: This is an extra check specific for KT1 addresses. A transaction to KT1 addresses can cost much more than a tx to t1 address (due to more gas and storage consumption). We would like to avoid such transactions with 0.1 xtz fees for 0.0001 payment amount for example. This will never happen with tz1 addresses but is possible for a costly KT1 contract.

src/pay/batch_payer.py
```diff
@@ -432,6 +444,8 @@ def attempt_single_batch(self, payment_records, op_counter, dry_run=None):
                 payment_item.paid = PaymentStatus.DONE
```
denver-s, line 444:
```suggestion
                payment_item.paid = PaymentStatus.AVOIDED
```

amzid: There was a discussion before concerning this and it was agreed back then to leave it to done so that the user is not confused. I would not change it in this PR, but you could open an issue to discuss this if you would like to.

amzid (approved):

src/pay/batch_payer.py
denver-s: The check for that is `if total_fee > FEE_LIMIT_CONTRACTS` right? Your edit added `if total_fee > pymnt_amnt` which is the same as done aftewards.

In my edits, KT1 did this: 
- total_fee = tx_fee + burn_fee
- check total_fee > FEE_LIMIT_CONTRACTS
- pymnt_amount = orig_amount - burn_fee

After that, for both KT1 and tz1 addresses, check that:

`pymnt_amnt = max(pymnt_amnt - tx_fee, 0)` which is the same as:

`pymnt_amnt = max(orgin_amount - burn_fee - tx_fee, 0)` or `pymnt_amnt = max(orgin_amount - total_fee, 0)`.

So I think the safety check is already done aftewards. Am I wrong?

amzid: With your code, this check would happen only if `delegator_pays_xfer_fee` is set to `True`, right? We would like to make sure that this check happens for KT1 accounts even if `delegator_pays_xfer_fee` is set to `False` if `KT1_FEE_SAFETY_CHECK` is `True`

denver-s: Now I get it, thanks. Fine for me.

PR #462, opened 2021-07-22
docker tag management

nicolasochem:
So far, we have only been pushing the "latest" tag.

We need to tag every docker image differently (release tags, commit tags)
to prevent unwanted upgrades of trd container.

We will still have a "latest" tag but we wouldn't recommend using it.

We are using the method described in github documentation:

https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-docker-hub

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,8 +26,9 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: trdo/tezos-reward-distributor
+
       - name: Push to Docker Hub
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v2
         with:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,8 +21,14 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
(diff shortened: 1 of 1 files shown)


jdsika:
@vkresch I am not experienced with docker. Explanation makes sense to me but could you approve, please?

.github/workflows/docker.yml
```diff
@@ -21,8 +21,14 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: trdo/tezos-reward-distributor
       - name: Push to Docker Hub
```
vkresch, line 29: Add one blank line above just for aesthetic reason ;)

.github/workflows/docker.yml
```diff
@@ -21,8 +21,14 @@ jobs:
...
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v5
```
vkresch, line 30: Does docker/build-push-action has already a version 5? On their [github page](https://github.com/docker/build-push-action) they still use v2 (see [example](https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md)). Or am I missing something?

vkresch (dismissed):
Looks good. Does docker/build-push-action has a version 5?

.github/workflows/docker.yml
nicolasochem: My bad, I was looking at the wrong action.

jdsika (approved):

PR #481, opened 2021-08-10
Support all tezos curves

utdrmac:
Work Effort: 2hr


teddyknox (approved):
Thanks!!

jdsika:
@teddyknox did you test the code?

jdsika:
@utdrmac lint is failing now - how did you test it? Can't we unit test this?

utdrmac:
> @utdrmac lint is failing now - how did you test it? Can't we unit test this?

I tested it with a standalone script. The issue with unit testing it is that you would need a revealed tz1/tz2/tz3 address on whichever chain you're testing against since the parent function makes an RPC call to fetch the public key.

Issue #496, opened 2021-08-23
TRD does not properly parse tzstats /tip


utdrmac:
Most recent cycle output (some output redacted for brevity):

```
2021-08-22 20:18:47,145 - producer  - DEBUG - Requesting https://api.tzstats.com/explorer/tip
2021-08-22 20:18:47,627 - producer  - DEBUG - Response from tzstats is: {'name': 'Tezos', 'network': 'Mainnet', 'symbol': 'XTZ', 'chain_id': 'NetXdQprcVkpaWU', 'genesis_time': '2018-06-30T16:07:32Z', 'block_hash': 'BM1qjCrd1SD4GCihSgaesDNxty8r7sCgJ3u9Hnn9fCfxpCJGodC', 'timestamp': '2021-08-22T20:18:36Z', 'height': 1630222, 'cycle': 393, 'total_accounts': 1620579, 'funded_accounts': 1464526, 'total_ops': 67404256, 'delegators': 132871, 'delegates': 394, 'rolls': 86527, 'roll_owners': 381, 'new_accounts_30d': 49600, 'cleared_accounts_30d': 8550, 'funded_accounts_30d': 52241, 'inflation_1y': 39500680.124225, 'inflation_rate_1y': 4.663378886902178, 'health': 95, 'deployments': [...], 'supply': {..}, 'status': {..}}
2021-08-22 20:18:47,628 - producer  - DEBUG - Checking for pending payments: payment_cycle <= current_cycle - (self.nw_config['NB_FREEZE_CYCLE'] + 1) - self.release_override
2021-08-22 20:18:47,628 - producer  - INFO - Checking for pending payments: checking 387 <= 392 - (5 + 1) - 0
2021-08-22 20:18:47,629 - producer  - INFO - No pending payments for cycle 387, current cycle is 392
2021-08-22 20:18:57,639 - producer  - DEBUG - Waiting until next cycle; 8195 blocks remaining
```

Note that the output from tzstats clearly says "cycle: 393", yet, 3 lines later, TRD believes the cycle is still 392. This is causing TRD to sleep beyond the new cycle start.

utdrmac:
This calculation, https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/e6ae93306de02cb11d12ab363df5d9f1fb7194d9/src/api/block_api.py#L25, returns -1 if level is the last level in a cycle which causes TRD to calculate sleep incorrectly.

[...]

PR #497, opened 2021-08-23
More accurate level-in-cycle calculation

utdrmac:
Cycle wakeup still having issues for some. This PR is continued investigations as to what is happening.

Do not merge until 'work-in-progress' label is removed.

**Work effort**: 4hr

### DIFF
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -4,6 +4,8 @@ EXIT_PAYMENT_TYPE = "exit"
 
 CURRENT_TESTNET = 'GRANADANET'
 
+FIRST_GRANADA_LEVEL = 1589249
+
 # Providers api prefix
 # Public RPC
 PUBLIC_NODE_URL = {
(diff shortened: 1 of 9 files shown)


utdrmac:
get_current_cycle and get_current_level were only ever called from one location in payment_producer.py. In some cases, the backend source was polled twice at the same API for the same information. I combined them together to save an API call.

utdrmac:
Fixes #496

nicolasochem (approved):
This checks out. The way continuous mode works is:
* figure out where we are in the cycle
* based on where we are, sleep for the expected duration

Granada code introduced a condition where performing this calculation at the very end of the cycle resulted in wrong sleep duration.

I am approving, but I did not try. I am not running continuous mode.

Few remarks:

* the issue that there is a granada level on mainnet but not on testnet has not been addressed; presumably TRD is broken on testnets because of that. There is already a TODO in the code
* it's been almost 11 cycles since granada so I think we can actually remove the "are we past granada?" check altogether.

But this is out of scope; ok to merge. Thank you very much for the thourough debugging, fix, and additional optimization of RPC calls.

PR #501, opened 2021-09-24
Re-run the tzkt tests

nicolasochem:
As I was working on some refactoring, I took a look at the tzkt
integration tests.

It appears that the tzkt tests are "cached" because they take one hour
to run.

Having discovered that, I then re-ran the tests. To do that, I deleted
the cached data in tests/integration/tzkt_data/*

I discovered that I actually broke the tests in #415

Previously, tzkt was paying out denunciation rewards. We have changed
this default behavior to not pay them.

The tzkt test compares the actual rewards from tzkt with the actual
rewards from rpc. But rpc actual rewards (calculated by the protocol)
take into account denunciation rewards. So we now have a discrepancy.

To solve it, I am hardcoding the reward value that we expect, for these
2 tests where there is a double baking gain.

This is a band-aid. We have several problems:

* tests should not be cached. What should be cached is the tzkt api
  calls so they return faster than the real things (we talked about that
  before)
* these tests should be generalized to tzkt, tzstats and api
* the cycles we are testing against are pre-granada. we should replace
  with more recent interesting cycles

Also there is a test mysteriously named "test staking balance issue"
which was also failing. I assume it is a reference to
[this](https://baking-bad.org/blog/2020/06/11/we-were-calculating-staking-balances-wrong/).

This test attempts to match total staking balance value from rpc and
tzkt in a cycle where the snapshot index was 15. An "assert_not_equal"
test was previously passing and is now failing, which means that some
values we expected not to match are now matching. Not sure how to
interpret this! I removed this statement.

@m-kus you wrote these tests and @utdrmac you are familiar with the
"staking balance issue", perhaps you can make more sense of this test
than I?

Other changes

* added comments so the next person is less confused than I was
* indent the json cached data

---
name: Pull Request
labels: 

---
IMPORTANT NOTICE:

* **Analysis**: If the described issue is a bug report, analyze the reasons resulting in this bug.

* **Solution**: Describe the proposed solution for the bug or feature.

* **Performed tests**: Describe the performed tests.

* **Check list**:

**Work effort**: Give your estimate of the work effort in hours. This might be adjusted or discussed by the other contributors in order to keep a fair rewarding process for the efforts.


vkresch:
@nicolasochem should this PR be merged before https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/509?

nicolasochem:
Yes

vkresch (approved):

PR #502, opened 2021-10-08
[WIP] Rework README + Documentation .rst to use VARIABLES

jdsika:
While researching for competitors I relaized how hard it was to find TRD, In addition all documentation needs to use variables for things like *HANGZHOU2NET*. Maybe [this](https://stackoverflow.com/questions/34006784/how-do-i-access-a-variable-in-sphinx-conf-py-from-my-rst-file) helpes?


jdsika:
This will be done later

PR #509, opened 2021-10-29
Refactoring + accusation payout toggle

nicolasochem:
When someone double bakes or endorses, anyone can submit an on-chain
proof of equivocation. The bakers that includes the proof gets rewarded
with half of the security deposit of the faulty baker.

It is important to note that equivocation has been very rare on Tezos so
this PR does not change TRD's behavior in the vast majority of cases.

Out of the discussions leading to PR #415 being merged, it has emerged
that TRD pays out accusation rewards inconsistently: RPC reward model
will rely on the on-chain calculation of rewards, which includes
accusation rewards. Tzstats does not include accusation rewards. Tzkt
used to, but #470 changed the tzkt code to not do it. We agreed to make
it configurable.

We introduce a new configuration option `payout_accusation_rewards`.
When using configure.py, the default choice is "False".

This PR has turned into a major refactoring: the logic to
calculate rewards used to be isolated per provider: tzkt, tzstats, rpc
would make their own calculations and return the reward amount via the
rewards model.

It makes more sense to have this calculation done at payment producer
level. To make this possible, I extended the model. While it was
previously retuning only one value `total_reward_amount`, it now returns
several additional values such as `equivocation_losses`,
`offline_losses`, `rewards_and_fees`, `accusation_rewards`.

Here are the new variables for the payment_producer object model:

```
self.delegator_balance_dict = delegator_balance_dict
self.delegate_staking_balance = delegate_staking_balance
self.num_baking_rights = num_baking_rights
self.num_endorsing_rights = num_endorsing_rights
\# rewards that should have been earned, had the baker been online
self.offline_losses = offline_losses

\# total reward as recorded in-protocol
self.total_reward_amount = total_reward_amount

\# When using indexers, the total amount above can be itemized as follows:

\# * baking rewards, fees, revelation rewards
self.rewards_and_fees = rewards_and_fees

\# * losses from double baking/endorsing
self.equivocation_losses = equivocation_losses

\# * rewards from denunciating other people's double baking/endorsing
self.denunciation_rewards = denunciation_rewards
```

The provider code populates these fields but no longer performs
calculation of the reward amount to be split.

Instead, this calculation is performed by the below core logic which is
provider-independent:

```
if rewards_type.isEstimated():
    logger.info("Using estimated rewards for payouts calculations")
    total_estimated_block_reward = reward_model.num_baking_rights * block_reward
    total_estimated_endorsement_reward = reward_model.num_endorsing_rights * endorsement_reward
    computed_reward_amount = total_estimated_block_reward + total_estimated_endorsement_reward
elif rewards_type.isActual():
    logger.info("Using actual rewards for payouts calculations")
    if self.pay_denunciation_rewards:
        computed_reward_amount = reward_model.total_reward_amount
    else:
        # omit denunciation rewards
        computed_reward_amount = reward_model.rewards_and_fees - reward_model.equivocation_losses
elif rewards_type.isIdeal():
    logger.info("Using ideal rewards for payouts calculations")
    if self.pay_denunciation_rewards:
        computed_reward_amount = reward_model.total_reward_amount + reward_model.offline_losses
    else:
        # omit denunciation rewards and double baking loss
        computed_reward_amount = reward_model.rewards_and_fees + reward_model.offline_losses
```

The result of tris calculation is a new variable
`computed_reward_amount` which is then passed to the `calculate`
function as a new argument. Previously `calculate` was relying on
`total_reward_amount` value in the model.

The following is true:

```
total_reward_amount = rewards_and_fees + accusation_rewards - equivocation_losses
```

RPC can't distinguish between an accusation reward and a
baking reward unless you parse every block (which is the job of an
indexer). RPC provider will populate `total_reward_amount` but not `rewards_and_fees` or
`accusation_rewards`. It will leave these fields to "None". This
explains why we left `total_reward_amount` in the model. While it can be
derived by other fields with tzstats or tzkt, this is not possible with
RPC.

Extending the model required several changes to tests as well.
Specifically tzkt tests continue to be a pain. In PR #501 we had
hardcoded the accusation rewards. Now, with our extended model, we can
compare apples to apples again, and I removed these hardcoded values.

Configuration
-------------

I modified configure.py to set the new flag. All flags are currently
mandatory, so I did not change that, but anyone upgrading TRD will have
to modify their config, which is not great. What does the team think?
Should we default to false when the config flag is absent?

Testing
-------

I did some manual testing with the following baker:

```
baking_address: tz1WnfXMPaNTBmH7DBPwqCWs9cPDJdkGBTZ8
```

On cycle ~78~74. This baker was rewarded for denunciating a double baking on
this cycle. I observed that the rewards were higher when the new setting
was set to "True". I also observed that the rewards value was consistent
between tzkt and tzstats.


jdsika:
This looks like something for a free weekend.... thank you for this contribution. It most definitely does remove some inconsistencies!

jdsika:
Conflicts due to previous merge of fixed tzkt tests

jdsika:
@vkresch can you schedule time to work for TRD on Saturday 27th?

tests/utils.py
```diff
@@ -131,5 +131,13 @@ def mock_request_get(url, timeout):
                 {"kind": "contract", "contract": "tz1gtHbmBF3TSebsgJfJPvUB2e9x8EDeNm6V", "change": "14626175032"}
             ]
         })
+    if path == "/chains/main/blocks/178177/helpers/baking_rights":
```
vkresch, line 134: To which tz address does this path belong to so that we need to mock it? Does the real path also return an empty list? I think it might be helpful to have tests for accessing endorsing_rights and baking_rights.

src/calc/phased_payment_calculator.py
```diff
@@ -42,10 +42,10 @@ def __init__(self, founders_map, owners_map, service_fee_calculator, min_delegat
     # owners reward = owners payment = total reward - delegators reward
     # founders reward = delegators fee = total reward - delegators reward
     ####
-    def calculate(self, reward_provider_model):
+    def calculate(self, reward_provider_model, total_rwrd_amnt):
```
vkresch, line 45: Do we need to pass it as an argument? We could just do:
```python
total_rwrd_amnt = reward_provider_model.total_reward_amount
```
locally inside the function.

vkresch (changes requested):
In general the PR looks great. The documentation of the new feature is really good and detailed.
The grouping of the calculation to just one place instead of to each individual provider is a really good step and design choice. 

I had just some minor comments and questions. Also the PR needs to be rebased based on the latest master. I do not want to do it myself because I would then rewrite some commit history which only the author should do.

I also realized that we should use a [test coverage tool](https://coverage.readthedocs.io/en/6.2/) to detect which code is covered by our tests. But that's a story for another or next PR.

src/pay/payment_producer.py
```diff
@@ -258,37 +265,54 @@ def try_to_pay(self, pymnt_cycle, rewards_type):
                 return True
 
             # 1- get reward data
+            reward_model = self.reward_api.get_rewards_for_cycle_map(pymnt_cycle, rewards_type)
+
+            block_reward = network_config["BLOCK_REWARD"]
+            endorsement_reward = network_config["ENDORSEMENT_REWARD"]
+
+            # 2- compute reward amount to distribute based on configuration
```
vkresch, line 273: I think we should make a compute_reward method which kinda looks like this:
```python
computed_reward_amount = compute_reward(reward_model, network_config)
```
Which then we could unit test for every possible scenario easier.

tests/integration/test_tzkt_reward_api.py
```diff
@@ -31,14 +41,20 @@ def store_reward_model(address, cycle, suffix, model: RewardProviderModel):
...
         total_reward_amount=model.total_reward_amount,
+        rewards_and_fees=model.rewards_and_fees,
+        equivocation_losses=model.equivocation_losses,
+        offline_losses=model.offline_losses,
+        num_baking_rights=model.num_baking_rights,
+        num_endorsing_rights=model.num_endorsing_rights,
+        denunciation_rewards=model.denunciation_rewards,
         delegator_balance_dict={
-            k: v
+            k: {i: v[i] for i in v if i != 'current_balance'}
```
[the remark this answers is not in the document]
nicolasochem: yes, but with the new model, I can't just put the staking balance in there. There are more fields now. I need "everything but the current balance".

tests/utils.py
nicolasochem: This block 178177 is actually never being queried during tests so I removed it. But the one below (196609) is needed during test of `cycle=51&delegate=tz1gtHbmBF3TSebsgJfJPvUB2e9x8EDeNm6V` in `TestCalculatePhases.test_process_payouts`.

The real path does not return an empty list. In this case, the model get filled with number of bake/endorses at zero, but it does not matter since we are not testing estimated rewards.

tests/integration/test_tzkt_reward_api.py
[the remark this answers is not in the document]
vkresch: Ok got it. Thanks for clarification.

vkresch (approved):

vkresch:
I will create after the merge a follow up PR with a unit test for the computation function and the test coverage tool.

jdsika:
@nicolasochem you have to put the long stories in the PR and only do precise texts as commit messages. It is sheer impossible for me to merge this with a useful comment :(

Issue #504, opened 2021-10-13
TRD repays when disk is full


chiptus:
**Describe the bug**
in cycle 403, our hard disk got full and trd couldn't save the reports. instead of closing, for some reason it decided to repay the same cycle. it stopped when our payment wallet emptied. (lucky for us, we don't pay directly from the baker)

**Expected behavior**
Stop TRD, and/or fail to pay

```
2021-10-10 19:02:15,391 - producer  - DEBUG - Unknown error in payment producer loop: [Errno 28] No space left on device
Traceback (most recent call last):
  File "/mnt/tezos/trd/src/pay/payment_producer.py", line 195, in run
    result = self.try_to_pay(pymnt_cycle, self.rewards_type)
  File "/mnt/tezos/trd/src/pay/payment_producer.py", line 302, in try_to_pay
    raise e from e
  File "/mnt/tezos/trd/src/pay/payment_producer.py", line 289, in try_to_pay
    self.create_calculations_report(reward_logs, report_file_path, total_amount, rewards_type)
  File "/mnt/tezos/trd/src/pay/payment_producer.py", line 377, in create_calculations_report
    writer.writerow(array)
OSError: [Errno 28] No space left on device
2021-10-10 19:02:15,392 - producer  - ERROR - Unknown error in payment producer loop: [Errno 28] No space left on device, will try again.
2021-10-10 19:02:18,804 - consumer0 - ERROR - Error, request ->GET https://rpc.tzkt.io:443/mainnet/chains/main/blocks/1765545/operation_hashes<-, params ->None<-,
---
Error, response -><-
```

PR #510, opened 2021-11-28
Disk size check

vkresch:
Resolves https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/504

* **Analysis**: As described in the issue TRD retries payments when the disk is full because it does not handle this edge case properly.

* **Solution**: The solution proposed in this PR is to check during the TRD run the disk size and to shut down properly (with the transition `SHUT_DOWN_ON_ERROR`) if the free disk size of the system is less then 10% ([source](https://serverfault.com/questions/10475/how-much-space-should-you-leave-free-on-a-hard-disk)). If we check everytime in the running loop the disk size we always make sure that TRD has enough disk space independent of the many write function location as described by [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/504#issuecomment-945411517). 

I was also considering and trying out adding a new state between `PRODUCERS_READY` and `CONSUMERS_READY`:
image
but the actual while loop for writing is inside the  `PRODUCERS_READY` state.
The `PRODUCER_READY` state is an aggregated state of TRD thus in future PRs we should consider splitting up this state into multiple smaller states to come closer to a finite state machine with all the exception handlings.

* **Performed tests**: Implemented test_disk_full.py regression test. Tested on own system increasing the constant percentage threshold.

* **Check list**:

**Work effort**: 7h


jdsika:
I think there is some minimum free space on a system that sometimes locks out non admin processes before the disc is even full. Like e.g. 5% free space -> all programs fail but the system can still write something in order to be able to gracefull handle the nearing full disk?

vkresch:
@jdsika yes good point I also found a discussion [here](https://serverfault.com/questions/10475/how-much-space-should-you-leave-free-on-a-hard-disk). To be on the safe side let's make it 10%. I will update the constant accordingly.

jdsika:
@vkresch will this be ready today?

vkresch:
I will try finishing it today but latest tomorrow.

jdsika:
I have merged the monster PR. You can resolve conflicts here and finish!

utdrmac:
Why isn't this failure being checked?
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/db45c466f6598b06bd89fcbd412af93b7a7c5ab1/src/util/csv_payment_file_parser.py#L31

This seems like the easiest solution. If you can't write() to a file, abort TRD completely. Checking for available disk space is a good preemptive test, but the actual writing should still send admin notification and halt TRD if writing fails.

src/Constants.py
```diff
@@ -1,22 +1,22 @@
...
+CURRENT_TESTNET = '{}NET'.format(PROTOCOL_NAME)
 
 FIRST_GRANADA_LEVEL = 1589249
 
 # Providers api prefix
 # Public RPC
 PUBLIC_NODE_URL = {
     'MAINNET': 'https://mainnet-tezos.giganode.io',
-    CURRENT_TESTNET: 'https://granadanet.smartpy.io'
+    CURRENT_TESTNET: 'https://{}.smartpy.io'.format(CURRENT_TESTNET)
```
vkresch, line 13: This is the culprit. The url` https://granadanet.smartpy.io` returns 502 error. Means it has been taken down.

src/Constants.py
```diff
@@ -1,22 +1,22 @@
 from enum import Enum
 
 EXIT_PAYMENT_TYPE = "exit"
-
-CURRENT_TESTNET = 'GRANADANET'
+PROTOCOL_NAME = 'GRANADA'
```
vkresch, line 4: Question: I tried updating the protocol name to `HANGZHOU` but it seems that the [current_level](https://hangzhounet.smartpy.io/chains/main/blocks/head/header/) of this protocol is at 169123. Which is significant less then in granada [current_level](https://rpc.tzkt.io/granadanet/chains/main/blocks/head/header) with 765770. 

I am asking because then it fails the `test_calculatePhase0.py` test. Because the snapshot_level (266241) is bigger then the current_level (169123) in `HANGZHOU`. See this if check [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/rpc/rpc_reward_api.py#L482). How do we want to deal with that?

tests/unit/test_calculatePhase0.py
```diff
@@ -15,7 +15,7 @@ def test_calculate(self):
 
         nw = DEFAULT_NETWORK_CONFIG_MAP[CURRENT_TESTNET]
 
-        api = ProviderFactory(provider='prpc').newRewardApi(nw, BAKING_ADDRESS, '')
+        api = ProviderFactory(provider='tzkt').newRewardApi(nw, BAKING_ADDRESS, '')
```
vkresch, line 18: Because https://granadanet.smartpy.io return 502 I use tzkt api.

vkresch:
> This may not be where TRD is running, nor where the user is saving reports. This should check disk space at the path of --reports-dir as that is the failure issue. If reports cannot be saved, then TRD repeats forever.

@utdrmac yes in the issue the error occured in the payment loop when it try_to_pay and create_calculations_report. I was thinking it might be beneficial to interrupt the main loop as early as possible to avoid unnecessary calculations and api requests.

src/cli/client_manager.py
```diff
@@ -24,8 +23,6 @@ def __init__(self, node_endpoint, signer_endpoint) -> None:
             else:
                 self.node_endpoint = 'https://' + self.node_endpoint
                 logger.info("Node endpoint URL points to an SSL endpoint. Using HTTPS protocol prefix.")
-        if len(self.node_endpoint.split(':')) < 3:
-            self.node_endpoint += f':{TEZOS_RPC_PORT}'
```
jdsika, line 28: Why did you remove this?

tests/smoke/test_launch_states.py
```diff
@@ -25,9 +25,9 @@
 @pytest.fixture
 def args():
     # Test with PRPC node
-    args = Args(initial_cycle=10, reward_data_provider='tzkt', api_base_url='https://api.carthage.tzkt.io/v1/')
+    args = Args(initial_cycle=10, reward_data_provider='tzkt', api_base_url='https://api.hangzhounet.tzkt.io/v1/')
```
jdsika, line 28: This needs to be the URL from constants.py ? it was hangzhou2net and I changed it

tests/smoke/test_launch_states.py
```diff
@@ -25,9 +25,9 @@
...
     args.network = 'EDO2NET'
```
jdsika, line 29: Needs change. I think we should run all tests here on mainnet and not testnet, right?

tests/smoke/test_tzkt_api.py
```diff
@@ -58,5 +58,5 @@ def test_base_url(ConfigParser, args):
         )
     )
     args.initial_cycle = 100
-    args.api_base_url = 'https://api.carthage.tzkt.io/v1/'
+    args.api_base_url = 'https://api.tzkt.io/v1/'
```
jdsika, line 61: This is the mainnet URL and above is the testnet URL from giganode. this doesn't seem to fit

jdsika:
Are you checking the disk space of --reports-dir as well?

tests/unit/test_calculatePhase0.py
jdsika: I updated the the testnet URL

tests/regression/test_disk_full.py
```diff
@@ -0,0 +1,166 @@
...
+)
+
+
+@pytest.fixture
+def args():
+    # Test with PRPC node
+    args = Args(
+        initial_cycle=10,
+        reward_data_provider="tzkt",
+        api_base_url="https://api.tzkt.io/v1/",
```
jdsika, line 37: I don't want any URLs in the tests. They should all be taken from constants.py! We are also again mixing testnet and mainnet. I would vote to do the tests only on mainner and not on testnet if there are especially tests that invole a specific block level etc

src/cli/client_manager.py
vkresch: The tezos rpc port url does not work anymore. Thus I removed it.

nicolasochem:
I have the same comment than @jdsika , we need to make sure that the partition used for reports_dir is the one checked for usage. [This](https://stackoverflow.com/a/12327880/207209) looks like a good method to do it in python.

src/Constants.py
jdsika: We will make tests with mainnet in order to avoid that trouble

jdsika:
@denver-s I need your help here quickly:

```
DEBUG    verbose:rpc_reward_api.py:245 [do_rpc_request] Requesting URL https://mainnet-tezos.giganode.io/chains/main/blocks/252928/context/delegates/tz1RpNqAzQNaZuxeLSbfEzVEzUWssddsL6Kw
ERROR    main.rpc_reward_api:rpc_reward_api.py:427 [get_d_d_b] RPC API Error: [do_rpc_request] Requesting URL 'https://mainnet-tezos.giganode.io/chains/main/blocks/252928/context/delegates/tz1RpNqAzQNaZuxeLSbfEzVEzUWssddsL6Kw' failed (500)
=========================== short test summary info ============================
FAILED tests/unit/test_calculatePhase0.py::TestCalculatePhase0::test_calculate
```

You added this baker address but it does not exist. I changed the network to mainnet as the testnetwork does always reboot and start at level 0. I am not sure what the test intends to do and would like to ask you to add a proper baker address that will not fail with future protocol upgrades. Thank you!

denver-s:
@jdsika Previously, it was using [another testnet addres](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/53693ab0942357fb9a88164d16c1dbd7197628c6/tests/unit/test_calculatePhase0.py) so I just changed to a Granadanet address. I am sure that it's better to use a mainnet address to avoid such errors.

jdsika (approved):

PR #529, opened 2021-12-20
Email plugin add features

ericlavoie:
---
name: Email plugin add features
about: Adding a few features to the email plugin
labels: 

---
IMPORTANT NOTICE:

* **Analysis**: Adding the option to not login when sending notification, if the recipients are not external, there is no need to login, and provide a user and password (in a windows/exchange environment writing in clear a password, opens access to internal systems, pull info, login on some systems etc etc., since accounts are enterprise wide, not restricted to the email servers) Having the option to relay internally is safer IMO.
* Also added an option to set a display name for the smtp sender.

* **Implementation**:  Modified email.py to add the 2 parameters, and changes sanity check to take into account user is no longer required as a parameter if using nologin, and use the old check if nologin is not set. Also set default for nolgin to false if it is not set.
* 2 new optional to be backward compatible existing configuration will not be impacted, they are
- smtp_nologin type bool True/False (optional)
- smtp_sender_name type string (optional)

* **Performed tests**: 
- Tested with none of the setting in the file, behave as normal.
- Tested with smtp_nologin false or true, behave as expected.
- Tested smtp_sender_name set or not set, behaved as expected.

* **Documentation**: 
* Updated plugins.md with the new information, explaining how using smtp_nologin is useful to send when no relay to an external destination is required, it removes the need to put a password in the clear in the config file.

* **Check list**:

**Work effort**: 3 hrs


jdsika:
@utdrmac I resolved the conflict and would ask you to give a review. You did create the email plugin intitally?

jdsika:
@vkresch I am not sure if I understand the formatter correctly. I thought it automatically formats the things I screw up in the CI pipeline or do I have to trigger something locally?

vkresch:
It is set to check if the black format was applied if not you need to run locally `black .`. You can install black via `pip install black`.

jdsika:
Auto format would be nice on commit?

ericlavoie:
> @utdrmac I resolved the conflict and would ask you to give a review. You did create the email plugin intitally?

I am not the creator of this plugin, This enhancements that I needed, I have extensive infrastructure experience especially emails.

jdsika:
Not doubting your stuff. Just the 4 eye principle should be kept. The lock file PR reminded me about that ...

ericlavoie:
> Not doubting your stuff. Just the 4 eye principle should be kept. The lock file PR reminded me about that ...

I expect it, I have just started with your code base, so review accordingly.
This one is more contained but still should be reviewed by more people.
My response was mainly from your question, and giving you a bit more context of my experience.
My experience with TRD codebase is still green.

jdsika:
It is REALLY cool that you are on board and so active :)

src/plugins/email.py
```diff
@@ -8,6 +8,7 @@
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 from email.utils import formatdate
+from email.utils import formataddr
```
utdrmac, line 11: imports from same lib can be combined to single line
`from email.utils import fromatdate, formataddr`

utdrmac (approved):

ericlavoie:
> It is REALLY cool that you are on board and so active :)

I'm having fun working with it, we had our own back end code, being alone coding with no review, it's not as fun :)
Keep in mind fir a while until I'm familiar, check if my submission misses some impacted section of the code.

src/plugins/email.py
ericlavoie: Agreed, modifying it

src/plugins/email.py
```diff
@@ -103,15 +108,31 @@ def validateConfig(self):
         self.sender = self.cfg["smtp_sender"]
         self.user = self.cfg["smtp_user"]
         self.password = self.cfg["smtp_pass"]
+        if "smtp_nologin" not in self.cfg:
+            self.cfg["smtp_nologin"] = False
+
+        self.nologin = self.cfg["smtp_nologin"]
+
+        if "smtp_sender_name" in self.cfg["smtp_sender_name"]:
```
jdsika, line 116: Is here the "not" missing? You are overriding when it is present?

ericlavoie: Not sure I understand, if the key exists I use it.

ericlavoie: Oh crap I see it, let me check

ericlavoie: 2 errors, not needed, and sit should not be the key, sicne it would throw an error if not set, check if the key is not in self.cfg, if not set it to None

ericlavoie:
Ran a few more tests, without the 2 new parameters set, it worked as expected, set one only test just in case, then tested setting just the other one, and it works as expected,

jdsika:
It would be best to actually write those tests - I know some have been lazy wrtiting tests but it catched those little issues. Would you be willing to extend the tests here as well. You can get support from @vkresch if you don't really know how to do it

jdsika (approved):

jdsika:
I will merge this to keep things going. Maybe you can add the tests another time

ericlavoie:
> It would be best to actually write those tests - I know some have been lazy wrtiting tests but it catched those little issues. Would you be willing to extend the tests here as well. You can get support from @vkresch if you don't really know how to do it

I would not mind at all, I probably would need help to figure out how you set them in your test suite.

PR #540, opened 2022-01-04
introduce adjusted early payout mode

nicolasochem:
This allows bakers to pay out cycles as soon as they are computed, 5
cycles before they actually run, without losing money.

This is done with a two phase reward calculation:

* phase 1 calculates estimated rewards based on baking and endorsing
  rights, and pays them out
* phase 2 recalculates rewards after the cycle actually runs. The
  difference between what should have been paid and what was paid out is
  called the overestimate. TRD tries to substract it as an adjustment
  during the ongoing payout.

I urge you to judge this feature not by its complexity, but by what it
brings to the table. It definitely makes the code more complex, but it
allows newer bakers to pay out rewards early (6-9 days after
delegating). This will make delegation and self-custody more popular.

Feature
=======

This PR does not change how TRD works when using a release_override of 0
(pay rewards after they unfreeze) or -5 (pay rewards after cycle runs).
So the vast majority of users will not use this new code.

To activate adjusted early payout, set `release_override` to -11. Any
time you try to pay a cycle that has not run yet, it will pay estimated
rewards and later substract an adjustment.

We removed "estimated" payout mode since anyone wishing to pay early
will prefer adjusted early payout. The valid payout modes are now
"actual" and "ideal". When used in combination with a release_override
of -11, it will pay estimated payouts and will later adjust to the
wanted "actual" or "ideal" amount.

This does not always work: it assumes delegator keeps delegating. When a
delegator leaves or drains their account, TRD may not be able to claim
back all dues. In this case, it will just claim back as much as possible
and pay nothing (adjustment will be equal to reward amount, so payout
will be zero).

It also works in reverse. In most situations the actual/ideal reward is
lower than the estimation, because of missed bakes and endorsements from
the baker and other bakers which result in lower rewards. But
actual/ideal also pay out fees and stolen block rewards, so it can
occasionally happen that the adjustment is positive and the rewards are
more than expected.

Regular payouts are the most precise and still the default
recommendation. Using adjusted early payouts still exposes the baker to
reduced earnings, but less so than just paying expected rewards. It can
be a way for bakers to differenciate.

Why not make this behavior configurable with a toggle? To avoid
configuration complexity, it is better to assume that people will want
this feature whenever they pay early. Plus, "ideal" vs "actual" is
now relevant when doing early payouts, so a toggle would not be enough
anyway. It's best to make the behavior implicit.

Calculation CSV file
====================

We introduce 3 new columns: overestimate, adjustment, and adjusted
amount.

Outside of the early payout mode (release_override 0 and -5), all three
columns will show zero value and the rest of the colums will be
identical to master branch today.

In adjusted early payout, the columns are populated as follows:

* "overestimate" is the difference between what we paid and what we
  should have paid. If we paid too much, it is positive. When the payout
  just happened, the values in the column show as "pending" as we do not
  know yet the value, until the cycle runs.
* "adjustment" is the value substracted from the payout amount due to a
  previous overestimate. It is equal to the negative of the
  overestimate of the just completed cycle (which is in a different csv
  file 6 cycles earlier!), or the full payout amount if the overestimate
  is above. In the latter case, it means we are not recovering all
  losses.
* adjusted_amount is the sum of the payout amount for the cycle being
  paid and the adjustment (generally a negative value). If the
  overestimate is greater than the payout amount, this value will be
  zero.  Adjusted_amount is the value actually being paid to the
  delegator.

This extra data makes it possible for the baker and their delegates to
interpret payout values as needed.

Logging
=======

The log displays evidence of recalculation of completed cycle (here 436)
and displays total adjustment (82,125,609 mutez) and the unrecovered
amount (15,201,928 mutez).
```
2021-12-31 11:17:15,670 - producer  - INFO - Checking for pending payments: checking 442 <= 437 - (5 + 1) - -11
2021-12-31 11:17:15,671 - producer  - WARNING - Please note that you are doing payouts for future rewards!!! These rewards are not earned yet, they are an estimation.
2021-12-31 11:17:15,671 - producer  - INFO - Payment cycle is 442
2021-12-31 11:17:15,672 - producer  - INFO - Checking for potential adjustment for recently completed cycle 436.
2021-12-31 11:17:15,672 - producer  - INFO - TRD ran for cycle: 436, calculating adjustments.
2021-12-31 11:17:15,697 - producer  - INFO - Computing rewards for cycle 436.
2021-12-31 11:17:16,740 - producer  - INFO - Using actual rewards for payouts calculations
2021-12-31 11:17:16,741 - producer  - INFO - Total rewards before processing is 129,163,176 mutez.
2021-12-31 11:17:16,745 - producer  - INFO - Total rewards after processing is 129,163,176 mutez.
2021-12-31 11:17:16,745 - producer  - INFO - Sum of amounts allocated to delegators is 129,163,170 mutez
2021-12-31 11:17:16,746 - producer  - INFO - Difference between total rewards and sum of amounts allocated to delegators is 6 mutez. This is due to floating point arithmetic. (max allowed diff is 10)
2021-12-31 11:17:16,746 - producer  - INFO - We overestimated payout for cycle 436 by 97,327,535.0 mutez.
2021-12-31 11:17:16,804 - producer  - INFO - Calculation report is created at ''
2021-12-31 11:17:16,804 - producer  - INFO - Computing rewards for cycle 442.
2021-12-31 11:17:17,969 - producer  - INFO - Using estimated rewards for payouts calculations
2021-12-31 11:17:17,971 - producer  - INFO - Total rewards before processing is 149,609,375 mutez.
2021-12-31 11:17:17,975 - producer  - INFO - Total rewards after processing is 149,609,375 mutez.
2021-12-31 11:17:17,975 - producer  - INFO - Total adjustment for past early payout is -82,125,609.0 mutez.
2021-12-31 11:17:17,975 - producer  - INFO - Adjusted total rewards is 67,483,766.0 mutez.
2021-12-31 11:17:17,975 - producer  - INFO - Sum of amounts allocated to delegators is 67,483,757.0 mutez
2021-12-31 11:17:17,975 - producer  - INFO - Difference between total rewards and sum of amounts allocated to delegators is 9.0 mutez. This is due to floating point arithmetic. (max allowed diff is 10)
2021-12-31 11:17:17,976 - producer  - INFO - After early payout of cycle 436, 15,201,928.0 mutez were not recovered.
2021-12-31 11:17:18,004 - producer  - INFO - Calculation report is created at ''
2021-12-31 11:17:19,006 - consumer0 - INFO - Starting payments for cycle 442
```

When adjusted early payout is not used, the logs show usual information.

Implementation
==============

A new function in payment_producer called `recompute_payouts` is called
in case of advance payouts. It calculates overestimate and returns an
`adjustment` map containing the amount to recover by address. This map
is passed to the calculation logic, which attempts to recover the
amounts.

`recompute_payouts` also modifies the calculation csv file of a previous
payout to record overestimate once computed. We add the new utility
`csv_calculation_file_parser` which can read and write this file. We
verified that reading and writing this file results in the exact same
file.

This utility is also abole to read an "old style" computation csv file
that does not contain the three new columns. They will be added at
subsequent write.

`compute_payouts` function has been extended to also call the reward api
model and the `calculate` function. This used to be done in
`try_to_pay`. `recompute_payouts` calls `compute_payouts` whenever the
overestimate is absent from the file. This results in reduced
duplication of data.

The actual payouts use the `adjusted_amount` value of the reward log
instead of `amount` as previously. We grepped the code for all instances
of `rl.amount` to make sure this happens in every case. In particular,
in case of not empty founders and owners map, or payout redirection, we
make sure that the "merge" still works and adjusted amounts are being
merged.

Tenderbake impact
=================

Tenderbake affects the baking and endorsing rewards computation, and it
also deprecates the unfreezing period. But it does not affect
the new mechanism introduced here. In particular, there will still be an
estimated amount, ideal and actual rewards. Merging this does not affect
the tenderbake workload.

Testing
=======

We tested this against an actual baker that pays estimated rewards. We
also tested the non-early payout. We compared csv files before/after and
verified that the adjustments match with the overestimates and that the
amounts being paid out are indeed adjusted. We ensured that the csv
files show the same values than master when used in "normal" mode.

The way adjusted rewards are calculated seem to introduce floating point
variance of 1 mutez. We modified the tzkt integration test to be
tolerant of this 1 tez difference, in order to make it pass.

Documentation
=============

We added a documentation page called "payout timing" to explain the
release_override parameter.

Update:
* Contributor: nicolasochem, Effort=compensated
* Reviewer: jdsika, Effort=4h
* Reviewer: ericlavoie, Effort=1h


docs/payouttiming.rst
```diff
@@ -0,0 +1,92 @@
+.. _payout_timing:
```
ericlavoie, line 1: Great work for this, it brings way more clarity that is needed.

decentralgabe:
First, awesome work. This is super interesting, and great clear explanation.

> This does not always work: it assumes delegator keeps delegating. When a
delegator leaves or drains their account, TRD may not be able to claim
back all dues. In this case, it will just claim back as much as possible
and pay nothing (adjustment will be equal to reward amount, so payout
will be zero).

This is the biggest risk and should be in big red bold letters. This is reason enough to put it behind an explicit configuration option.

At the same time I'd love to see a TZIP that allows delegators to 'commit' to a baker for a certain amount of cycles.

nicolasochem:
> This is the biggest risk and should be in big red bold letters. This is reason enough to put it behind an explicit configuration option.

Thanks for looking into the PR. The text is already in bold. I think a configuration option is redundant as you already have to choose `--release_override -11` to activate this behavior.

ericlavoie:
I agree, it is implicit, this is why I liked the new section in the docs you created explaining fully the options when using release override at -11, and -5. No need of a parameter to a parameter.

src/calc/calculate_phaseMerge.py
```diff
@@ -35,6 +35,10 @@ def calculate(self, reward_logs):
                 total_current_balance = sum([rl.current_balance for rl in rl_list])
                 total_ratio = sum([rl.ratio for rl in rl_list])
                 total_payment_amount = sum([rl.amount for rl in rl_list])
+                total_adjusted_payment_amount = sum(
```
ericlavoie, line 38: you seem to be replacing amount and payment amount by an entry called adjusted amount, this seemed focus by the adjustment mode logic, but does not fit the logic when amounts are not in adjusted mode.
IMO adjusted label should be to specific adjusted mode code, if at we are at -5 and not -11 it makes no sense when parsing the code.

src/calc/calculate_phase_final.py
```diff
@@ -29,7 +29,19 @@ def calculate(self, reward_data5, total_amount):
...
+            if adjustments and rl.address in adjustments.keys():
+                rl.adjustment = max(-adjustments[rl.address], -rl.amount)
+            else:
+                rl.adjustment = 0
+            rl.adjusted_amount = int(
+                Decimal((rl.ratio * total_amount) + rl.adjustment).to_integral_value(
+                    rounding=ROUND_HALF_DOWN
+                )
+            )
+            rl.payable = (
```
ericlavoie, line 41: the condition adjusted_amount > 0 is this correct for -5? sounds like this would be at 0 and it would skip and it should not (compared to the original logic)

nicolasochem: We are using this formula: `adjusted_amount = amount + adjustment`

In case of `-5`, `rl.adjustment` would be zero so `rl.adjusted_amount` is equal to `rl.amount`. So it's only marked payable if it's positive.

nicolasochem:
@ericlavoie here is an extract of a csv file for adjusted amount calculation for a recent payment on a baker that is using this mode

```
| address                              | type            |  staked_balance | current_balance |  ratio | fee_ratio |      amount | fee_amount | fee_rate | overestimate | adjustment | adjusted_amount | payable | skipped | atphase | desc                      | payment_address                      | rewards_type |
| ------------------------------------ | --------------- | --------------- | --------------- | ------ | --------- | ----------- | ---------- | -------- | ------------ | ---------- | --------------- | ------- | ------- | ------- | ------------------------- | ------------------------------------ | ------------ |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | B               | 325,693,052,845 |               1 | 1.000… |    0.000… | 109,267,268 |          0 |     0.00 | pending      | -1,260,986 |     108,006,277 |   False |   False |      -1 | Baker                     |                                      | E            |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |  42,427,338,076 |  42,467,613,316 | 0.136… |    0.007… |  14,908,774 |    784,672 |     0.05 | pending      |   -174,605 |      14,734,169 |    True |   False |      -1 |                           | tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |  20,872,161,296 |  20,891,974,330 | 0.067… |    0.004… |   7,334,383 |    386,020 |     0.05 | pending      |    -85,897 |       7,248,486 |    True |   False |      -1 |                           | tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
| KT1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |  17,238,355,930 |  17,254,718,997 | 0.055… |    0.003… |   6,057,480 |    318,815 |     0.05 | pending      |    -70,943 |       5,986,537 |    True |   False |      -1 |                           | KT1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |  15,151,415,645 |  15,165,798,008 | 0.049… |    0.003… |   5,324,139 |    280,218 |     0.05 | pending      |    -62,354 |       5,261,785 |    True |   False |      -1 |                           | tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |  12,941,175,975 |  12,953,460,169 | 0.042… |    0.002… |   4,547,470 |    239,341 |     0.05 | pending      |    -53,258 |       4,494,212 |    True |   False |      -1 |                           | tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |   8,099,425,603 |   8,107,113,552 | 0.026… |    0.001… |   2,846,101 |    149,795 |     0.05 | pending      |    -33,332 |       2,812,769 |    True |   False |      -1 |                           | tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
| tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | D               |   8,063,483,577 |   8,071,137,407 | 0.026… |    0.001… |   2,833,471 |    149,130 |     0.05 | pending      |    -33,184 |       2,800,287 |    True |   False |      -1 |                           | tz1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | E            |
```

src/calc/calculate_phase_final.py
ericlavoie: Thanks for the precision

ericlavoie: Thanks for the precision

ericlavoie:
This looks good to me, your work is detailed and extremely well documented, I asked others more knowledgeable to this code base to weight in.

jdsika:
@nicolasochem I am troubled by this here:

> We removed "estimated" payout mode since anyone wishing to pay early
> will prefer adjusted early payout. The valid payout modes are now
> "actual" and "ideal". When used in combination with a release_override
> of -11, it will pay estimated payouts and will later adjust to the
> wanted "actual" or "ideal" amount.

I cannot say if people do want to adjust their payouts. To my knowledge there are bakers that use the estimated payouts mode and I think they have no desire to adjust it or change the mode. We would need to reach out to them when we remove the feature. What made you think you should remove the mode?

nicolasochem:
> I cannot say if people do want to adjust their payouts. To my knowledge there are bakers that use the estimated payouts mode and I think they have no desire to adjust it or change the mode. We would need to reach out to them when we remove the feature. What made you think you should remove the mode?

Yes, do you know who these bakers are and can you reach out? I would think that they would happily switch to adjusted early payouts: their payouts will be more accurate and they will lose less tez.

In the current state of the PR, anyone using "estimated" payout will have to switch their config to either "actual" or "ideal". If they leave release_override to -11, they will still pay estimated rewards, but the future payouts will be adjusted to match "actual" or "ideal" rewards (based on rewards_type) once we are able to compute them.

2 reasons why I opted to remove the mode:

* no reason to pay estimated rewards if this adjustment mechanism exists
* if you choose to pay adjusted early payouts, you still have to pick "ideal" or "actual" in order to pay the adjustment

What's the alternative then? We could have 4 options `payout_type: [ actual | ideal | estimated_adjusted_to_actual | estimated_adjusted_to_ideal]` but that's ugly.

Or 2 configuration parameters:

```
payout_type: [ actual | ideal | estimated ]
adjustment_type: [ actual | ideal ] # only settable when payout_type is 'estimated'
```

This is ugly too. In the PR currently, when release_override is -11, we pay estimated rewards. No need for a "mode". This is simpler and less confusing. Don't forget that a vast majority of bakers today don't care because they don't pay in advace.

jdsika:
@nicolasochem

> Yes, do you know who these bakers are and can you reach out? I would think that they would happily switch to adjusted early payouts: their payouts will be more accurate and they will lose less tez.

The problem is that IMO the payout mode is a "customer promise" by the baker - if I personally think it is useful or if I would do it like that is not up to discussion. You should always consult the other maintainer before you intend to cancel features of TRD. I don't see myself in the position to decide for the community a change like that. I have to think how we can proceed here as I don't want the new feature to not be merged. 

@utdrmac Do you see a way of knowing or signaling bakers that use this mode?

nicolasochem:
Maybe it's not clear. The feature is still here. You can still pay 11 cycles in advance. The way to configure it is changing. Also, after you run it for a while, it starts adjusting payouts so the baker loses less money. I didn't make the adjustment optional since I see no good reason anyone would disable it.

ericlavoie:
> Maybe it's not clear. The feature is still here. You can still pay 11 cycles in advance. The way to configure it is changing. Also, after you run it for a while, it starts adjusting payouts so the baker loses less money. I didn't make the adjustment optional since I see no good reason anyone would disable it.

The issue is one of backward compatibility, anyone updating their code, that use estimated, this will now fail, like we all know not eveyone keeps up to date.  Question we collect stats data from bakers (if they opted in) can we see how many are using this parameter?

The other option is to leave the parameter (and mark it deprecated) and handle the cases, I assume using estimated means someone is using a value of -6 to -11 for release_override. If that is the case to be backward compatible, if someone has set estimated, we can default to actual. Can this work and resolve the issue.

nicolasochem:
> The issue is one of backward compatibility, anyone updating their code, that use estimated, this will now fail, like we all know not eveyone keeps up to date. Question we collect stats data from bakers (if they opted in) can we see how many are using this parameter?
> 

I suppose so, though I'm not familiar with the stats collection infra.


> The other option is to leave the parameter (and mark it deprecated) and handle the cases, I assume using estimated means someone is using a value of -6 to -11 for release_override. If that is the case to be backward compatible, if someone has set estimated, we can default to actual. Can this work and resolve the issue.

Yes, it works, though the hidden behavior makes me uneasy. Instead, I just pushed a commit that simply fails TRD in error when rewards_type is set to estimated:

```
2022-01-20 14:42:55,231 - MainThread - ERROR - [Process Life Cycle completing With Failure] Error Details: 'ESTIMATED' is no longer a valid option for rewards payout type.          
Please see https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/540                                                                                
2022-01-20 14:42:55,233 - MainThread - INFO - TRD is shutting down...                           
```

People using this mode, who upgrade their TRD, will come here and gain all the necessary context. Is this acceptable?

ericlavoie:
May I suggest you lead with:
estimated reward has been deprecated, please see here [link] for details on how to configure a improved method.

ericlavoie:
Use the doc section that explains the new parameter

jdsika:
> Maybe it's not clear. The feature is still here. You can still pay 11 cycles in advance. The way to configure it is changing. Also, after you run it for a while, it starts adjusting payouts so the baker loses less money. I didn't make the adjustment optional since I see no good reason anyone would disable it.

The mode "estimated" was meant as a kind of baker insurrance. "However bad we are at baking - you get what was estimated". It was always clear that bakers can lose money with it. It is no unwanted behavior that needed fixing. We have to figure out how many people are troubled by depricating it. I asked BakingBad for help.

ericlavoie:
> > Maybe it's not clear. The feature is still here. You can still pay 11 cycles in advance. The way to configure it is changing. Also, after you run it for a while, it starts adjusting payouts so the baker loses less money. I didn't make the adjustment optional since I see no good reason anyone would disable it.
> 
> The mode "estimated" was meant as a kind of baker insurrance. "However bad we are at baking - you get what was estimated". It was always clear that bakers can lose money with it. It is no unwanted behavior that needed fixing. We have to figure out how many people are troubled by deprecating it. I asked BakingBad for help.

Ah ok, bakingbad has bakers insurance, meaning if you miss a block or endorsement, the delegate still get the expected reward. It's a different feature.

dmirgaleev:
@nicolasochem, that is a great idea, but it brings complexity for TRD, users, and auditors. Mixing up estimated rewards for one cycle with adjustment payouts for another makes it almost impossible to check if something went wrong.

nicolasochem:
> May I suggest you lead with: estimated reward has been deprecated, please see here [link] for details on how to configure a improved method.

Done in the last commit. Here is how it looks like now:

```
2022-01-21 08:24:30,888 - MainThread - ERROR - [Process Life Cycle completing With Failure] Error Details: Setting 'rewards_type' to 'estimated' is no longer supported.
Please see https://tezos-reward-distributor-organization.github.io/tezos-reward-distributor/payouttiming.html
for details on how to configure a improved method.
```

> The mode "estimated" was meant as a kind of baker insurrance. "However bad we are at baking - you get what was estimated". It was always clear that bakers can lose money with it.   I asked BakingBad for help.

If people want bakers insurance, they should use mode `ideal`. This mode pays out your baking and endorsement misses, but not the loss of income due to other baker's downtime. The only reason why it makes sense to use `estimated` is when you pay rewards upon assignment, before the cycle runs.

> It is no unwanted behavior that needed fixing.

The purpose of this PR is not to fix anything, it is to introduce a new feature to TRD. Since this new feature is better than previous estimated mode, I'm removing this mode. I can bring it back if that's what it takes to get this PR merged, but please consider the case that I am making.

> We have to figure out how many people are troubled by depricating it.

From your slack poll it would appear that most people are using actual. This is also my intuition.

> @nicolasochem, that is a great idea, but it brings complexity for TRD, users, and auditors. Mixing up estimated rewards for one cycle with adjustment payouts for another makes it almost impossible to check if something went wrong.

I agree, this is a more complex formula that what we have been using and if popular, it will require changes in baking bad audit mechanism and the insurance engine. If something goes wrong, the csv file is here to help you debug. The calculation method is not that complex and you can port it to your codebase: "after the cycle runs, if there is a difference between estimated and actual, attempt to adjust"

The larger point is that it allows people to safely pay 10 days after delegation instead of 30. It gives new bakers an edge if they so choose. It could spur renewed competition across bakers and renewed interest in staking.

ericlavoie:
The explanation of ideal for insurance case, answer my question, bad understanding on my part.

jdsika:
A very constructive comment from the slack:

Make the "adjusted payments" as an option that works with the "estimated" payout mode. This way we keep backward compatibility? What do you think?

dmirgaleev:
> I agree, this is a more complex formula that what we have been using and if popular, it will require changes in baking bad audit mechanism and the insurance engine. If something goes wrong, the csv file is here to help you debug. The calculation method is not that complex and you can port it to your codebase: "after the cycle runs, if there is a difference between estimated and actual, attempt to adjust"
> 
> The larger point is that it allows people to safely pay 10 days after delegation instead of 30. It gives new bakers an edge if they so choose. It could spur renewed competition across bakers and renewed interest in staking.

Unfortunately, without TRD csv files, sometimes it's not just complex to audit, but impossible to audit, because in one transaction can be mixed rewards for two cycles. And sometimes in that transaction have been added rewards for addresses from rules_map

nicolasochem:
> A very constructive comment from the slack:
> 
> Make the "adjusted payments" as an option that works with the "estimated" payout mode. This way we keep backward compatibility? What do you think?

I can add `early_payout_adjustment: true` in the default configuration, so this slack user will be able to set it to `false` when/if they become a public baker. I maintain that paying estimated rewards without adjustment makes bakers lose money on average.

But if I do this, everyone will have to change their configuration, because in the current state of TRD, there is no way to have a configuration field be optional. This will affect all TRD users. In the current state of my PR, no one is affected, because no one but me currently uses estimated mode. @jdsika are you sure you want a config toggle?

@dmirgaleev in one transaction, there are mixed rewards for 2 cycles, but they are still predictable based on a calculation. So it is not impossible to audit. If you have special rules in rules_map, you can't do an audit without access to the config, with or without my PR.

I think most delegators don't look closely at reward amounts. But they pay attention to the 21/37 day delay. Yes it's more complicated, but if it's popular, so what? If it's not popular, there is nothing you need to do, so no reason to stonewall the PR.

jdsika:
> I can add early_payout_adjustment: true in the default configuration, so this slack user will be able to set it to false when/if they become a public baker. I maintain that paying estimated rewards without adjustment makes bakers lose money on average.

I would opt for the default value to be `early_payout_adjustment: false` - check the config for the presence of the value at startup - the config is parsed anyway. If not present but it in with false. Do a notification for the user like "there is an additional option to adjust payments. Have a look at the documentation"

But let's think through the different combinations. 3 payout modes with the additional flag. How would all of them behave? 

I will ask the user if he REALLY wants the mode. I am very reluctant to add the complexity.

nicolasochem:
> I would opt for the default value to be `early_payout_adjustment: false` - check the config for the presence of the value at startup - the config is parsed anyway. If not present but it in with false. Do a notification for the user like "there is an additional option to adjust payments. Have a look at the documentation"

Unclear, are you saying we should support optional flags in the config? Today, it's not the case. If I add a flag, everyone has to add it to their config otherwise we throw an error. I think optional flags make sense in general and I may work on this next. I was hoping that we do this as part of the big config refactor where we move most CLI options to the config.

Also, trying to communicate with the user through the log is not great.

> But let's think through the different combinations. 3 payout modes with the additional flag. How would all of them behave? 

ok, let's do it :)

`rewards_type: actual` `early_payout_adjustment: false` `release_override:0/-5`: pay actual rewards (ignore new flag)
`rewards_type: actual` `early_payout_adjustment: true` `release_override:0/-5`: pay actual rewards (ignore new flag)
`rewards_type: actual` `early_payout_adjustment: false` `release_override:-11`: fail
`rewards_type: actual` `early_payout_adjustment: true` `release_override:-11`: fail
`rewards_type: ideal` `early_payout_adjustment: false` `release_override:0/-5`: pay ideal rewards(ignore new flag)
`rewards_type: ideal` `early_payout_adjustment: true` `release_override:0/-5`: pay ideal rewards (ignore new flag)
`rewards_type: ideal` `early_payout_adjustment: false` `release_override:-11`: fail
`rewards_type: ideal` `early_payout_adjustment: true` `release_override:-11`: fail
`rewards_type: estimated` `early_payout_adjustment: true` `release_override:0/-5`: pay actual rewards (or should it be ideal? do we need a third flag?)
`rewards_type: estimated` `early_payout_adjustment: false` `release_override:0/-5`: pay estimated rewards (even though we are able to calculate ideal/actual)
`rewards_type: estimated` `early_payout_adjustment: false` `release_override: -11`: pay estimated rewards(early_payout_adjustment setting is ignored)
`rewards_type: estimated` `early_payout_adjustment: true` `release_override:-11`: pay estimated rewards, adjust to actual (or should it be ideal? do we need a third flag?)


The new flag is either relevant or irrelevant based on the status of other flags. This is super ugly and not a single user will understand how it works.

My solution (currently implemented)

`rewards_type: actual` `release_override:0 or -5`: pay actual rewards
`rewards_type: actual` `release_override:-11`: pay estimated rewards, adjust to actual after the cycle runs
`rewards_type: ideal` `release_override:0 or -5`: pay ideal rewards
`rewards_type: ideal` `release_override:-11`: pay estimated rewards, adjust to ideal after the cycle runs

jdsika:
@dansan566 I want to quickly discuss the following:

> This does not always work: it assumes delegator keeps delegating. When a
> delegator leaves or drains their account, TRD may not be able to claim
> back all dues. In this case, it will just claim back as much as possible
> and pay nothing (adjustment will be equal to reward amount, so payout
> will be zero).

There is another "EdgeCase": The adjustment will also take the CURRENT fee which means:
* You payout at e.g. -11 with 10% fee
* You adjust payments later (when the cycle is complete) with a fee of 15%
* Result: The already "prepaid" cylce from -11 is then adjusted to 15% fee

How do you evalutate this. How do you consider the "risk" of overpaying clients that cannot be adjusted later?

jdsika:
@nicolasochem I agree that the flag is terrible...

> My solution (currently implemented)
> 
> rewards_type: actual release_override:0 or -5: pay actual rewards
> rewards_type: actual release_override:-11: pay estimated rewards, adjust to actual after the cycle runs
> rewards_type: ideal release_override:0 or -5: pay ideal rewards
> rewards_type: ideal release_override:-11: pay estimated rewards, adjust to ideal after the cycle runs

Just to avoid misunderstanding. You want the value ranges in between to be forbidden and your implementation throws e.g. an error when typing -10?

jdsika:
@dmirgaleev thank you for researching if bakers use the payout mode "estimated". The result from your side and the reactions in Slack resulted in currently 0 users. As we have a large change ahead with Tenderbake I would lean towards replacing the "estimated" option with a potentially valuable feature.
People delegating do often have a hard time understanding that they have to wait a long time for their rewards. Bakers might get a competitive advantage paing upfront and alos draw new users to the Tezos ecosystem. I think this is a good thing overall.

I fully understand your concerns about the audit! It is easy for me to say (as not being an auditor) but I believe the potential upside for the ecosystem that the feature brings might be worth the trouble.

@nicolasochem maybe when advertising the feature we could ask bakers to approach BakingBad and tell them about their plans to use this feature and help them to adapt their auditing?

dmirgaleev:
@jdsika I hope you're right. Although one baker has this feature, I can't say it's in high demand among the Tezos community. 
Also, I have to inform you, we'd be happy to implement such a feature for the Baking Bad delegator's dashboard (Payment Auditor), but it's technically impossible, unfortunately. Payout amounts will be provided as "Inaccurate" or "No data".
Here's how it looks like for the baker, implemented that feature: https://baking-bad.org/shared/61f2b83fa6632298f27ed6d9

dmirgaleev:
@jdsika when we started the Baking Bad project in 2018, there were a lot of payment strategies, but all bakers ended up with just one because it works great and everyone is happy with that, actually.

ericlavoie:
I tend to agree, this use case is ultra fringe at best, I don't see myself ever using it, too much risk. Paying before unfreezing makes sense at the earliest. And seeing that very few use it. Is it worth the complexity that will need to be maintained.

nicolasochem:
> Just to avoid misunderstanding. You want the value ranges in between to be forbidden and your implementation throws e.g. an error when typing -10?

No, the current implementation does not do that. I would like only 0, -5 and -11 to be allowed, yes. I didn't implement because it wouldn't work on test networks where PRESERVED_CYCLES is different, so it would not be -5/-11 then. This has to be tested against test networks, so maybe that can be a followup discussion.

> I tend to agree, this use case is ultra fringe at best, I don't see myself ever using it, too much risk. Paying before unfreezing makes sense at the earliest. And seeing that very few use it. Is it worth the complexity that will need to be maintained.

I will have to maintain this branch regardless of whether we merge it. Payout mode estimated is already in the code (not added by me) so there is already something to maintain (arguably there are less lines of code than what I want to merge).

"Ultra fringe" for today's bakers and delegators. How about everyone who stakes on coinbase/kraken because they get fast payout? Maybe this feature can draw them to self custody.

jdsika:
@nicolasochem can you add a warning at the startup of TRD that tells the user something like:

"You have configured TRD to payout estimated rewards X cycles ahead of their potential realization. There might be unforseen attack vectors like being unable to adjust rewards of delegators that leave your baker. Are you aware of the risks? y/N" -> users acknowledges.

nicolasochem:
@jdsika I can't add a Y/N dialog, unless I also implement a `--force` parameter for people running TRD in non-interactive mode (like me).

Why is an additional warning needed? There is already a warning:

```"Please note that you are doing payouts for future rewards!!! These rewards are not earned yet, they are an estimation."  ```

Today, if people pay estimated rewards from master branch, there is no adjustment, so this "attack vector" is much worse than in this PR.

I have added an extra line. Is this acceptable?

```
                             logger.warn(
                                 "Please note that you are doing payouts for future rewards!!! These rewards are not earned yet, they are an estimation."
+                                "TRD will attempt to adjust the amount after the cycle runs, but it may not work."
                             )
```

**Review by ericlavoie — APPROVED** [review]
Perfect

File: src/calc/phased_payment_calculator.py [hunk 4]
```diff
@@ -125,21 +125,43 @@ def calculate(self, reward_provider_model):
...
         )
-        logger.info("Total amount to pay is {:,} mutez".format(total_amount_to_pay))
+        if total_adjustments < 0:
+            logger.info(
+                "Total adjustment for past early payout is {:,} mutez.".format(
+                    total_adjustments
+                )
+            )
+            logger.info(
+                "Adjusted total rewards is {:,} mutez.".format(
```
jdsika, line 154: Let's do all loggings in tez, please

jdsika: I saw that this is used differently all over the code. I will clean this up later

src/configure.py
```diff
@@ -150,9 +150,7 @@ def onrewardstype(input):
         rt = RewardsType(input.lower())
         parser.set(REWARDS_TYPE, str(rt))
```
jdsika, line 151: @nicolasochem don't we have to drop the  `RewardsType:ESTIMATED` from `Constants.py` ???

jdsika: You tell people to only use the two supported types but if someone types "estimated" you allow it and catch the error later in the parser, right?

jdsika: aaaand you need the type because you set it later to ESTIMATED when cecking for [-11,-6] RELEASE_OVERRIDE. ok ok ok

src/pay/payment_producer.py
```diff
@@ -450,42 +555,67 @@ def try_to_pay(self, pymnt_cycle, rewards_type, network_config):
...
 
-            # 1- get reward data
-            reward_model = self.reward_api.get_rewards_for_cycle_map(
-                pymnt_cycle, rewards_type
-            )
+            adjustments = {}
+            early_payout = False
+            current_cycle_rewards_type = rewards_type
+            # 1- adjust past cycle if necessary
+            if self.release_override == -11 and pymnt_cycle >= current_cycle:
```
jdsika, line 562: @nicolasochem this seems odd. I think it needs to work in the range of [-11,-6] ?

src/configure.py
nicolasochem: Yes, correct. `ESTIMATED` is still used internally (and visible in calculations.csv) but I made it impossible to configure it in the config file.

src/pay/payment_producer.py
nicolasochem: @jdsika I have not tried in [-10, -6] range. I think we should limit the modes of opertion of TRD to 0, -5 and -11. I don't think anyone would use anything else.

With the current state of the code, you must use -11 for adjusted early payouts to kick in. Any other value in [-6, -10] will simply pay early payouts.

We should probably restrict the override value to 0, -5 and -11. I have not done it because these values would actually be different in testnets (because PRESERVED_CYCLE chain constant differs). Testnet is not in scope of the PR, but we should take a look at it otherwise.

jdsika (approved):

Issue #511, opened 2021-12-03
when RPC returns 4xx, terminate immediately, do not retry


nicolasochem:
There is a retry look in TRD RPC requests.
500 means that the server is unresponsive/down so retry is legitimate. But a 404 error means that the node does not have the requested data (it's not an archive node). In this scenario, we should exit in error, because a retry will not help.

vkresch:
Do we want to exit all 4xx codes or only the 404 code? Currently there is already a raise ApiProviderException for 404 status codes [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/rpc/rpc_reward_api.py#L374) and also a general raise for all codes which are not 200 [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/rpc/rpc_reward_api.py#L380).

nicolasochem:
Thanks for looking into this. I don't remember :(
It was probably not a 404, then.

Any 4xx code should raise an exception. We should only have retry loops for 5xx.

PR #545, opened 2022-01-15
Fix/terminate rpc request

vkresch:
Extending unit tests for the APIs. 
All the API requests are patched and do not do actual API requests.

I added a test for the 404 raise exception which shouldn't do a retry like described in this [issue](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/511). 
I also added a retry counter to avoid infinite loops. The max count of retries is 256 times which I found in the tzkt code and then moved to the Constants.py.

Resolves https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/511

**Work effort**: 5h

### DIFF
--- a/src/rpc/rpc_reward_api.py
+++ b/src/rpc/rpc_reward_api.py
@@ -208,16 +208,14 @@ class RpcRewardApiImpl(RewardApi):
             # necessary data to properly compute rewards
             raise e from e
 
-    def __get_response(self, address, request_function, response=None):
+    def __get_response(self, address, request, response=None):
         retry_count = 0
         while (not response) and (retry_count < MAX_SEQUENT_CALLS):
             retry_count += 1
             sleep(RPC_REQUEST_BUFFER_SECONDS)  # Be nice to public RPC
             try:
                 logger.info("Fetching address {:s} ...".format(address))
-                response = self.do_rpc_request(
-                    request_function, time_out=5
-                )
+                response = self.do_rpc_request(request, time_out=5)
             except requests.exceptions.RequestException as e:
                 # Catch HTTP-related errors and retry
                 logger.warning(
@@ -558,7 +556,9 @@ class RpcRewardApiImpl(RewardApi):
                 get_staking_balance_request = COMM_DELEGATE_BALANCE.format(
                     self.node_url, level_snapshot_block, delegator
                 )
-                d_info["staking_balance"] = int(self.__get_response(delegator, get_staking_balance_request))
+                d_info["staking_balance"] = int(
+                    self.__get_response(delegator, get_staking_balance_request)
+                )
 
                 sleep(
                     0.5
--- a/src/rpc/rpc_reward_api.py
+++ b/src/rpc/rpc_reward_api.py
@@ -208,6 +208,30 @@ class RpcRewardApiImpl(RewardApi):
             # necessary data to properly compute rewards
             raise e from e
 
+    def __get_response(self, address, request_function, response=None):
+        retry_count = 0
+        while (not response) and (retry_count < MAX_SEQUENT_CALLS):
+            retry_count += 1
+            sleep(RPC_REQUEST_BUFFER_SECONDS)  # Be nice to public RPC
+            try:
+                logger.info("Fetching address {:s} ...".format(address))
+                response = self.do_rpc_request(
+                    request_function, time_out=5
+                )
+            except requests.exceptions.RequestException as e:
+                # Catch HTTP-related errors and retry
+                logger.warning(
+                    "Failed with exception, will retry ({}) : {}".format(
+                        retry_count, str(e)
+                    )
+                )
+                sleep(RPC_RETRY_TIMEOUT_SECONDS)
+            except Exception as e:
+                # Anything else, raise up
+                raise e from e
+
+        return response
+
     def __get_baking_rights(self, cycle, level):
         """
         Returns list of baking rights for a given cycle.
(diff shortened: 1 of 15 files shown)


src/rpc/rpc_reward_api.py
```diff
@@ -376,8 +380,13 @@ def do_rpc_request(self, request, time_out=120):
                 "RPC URL '{}' not found. Is this node in archive mode?".format(request)
             )
 
-        # URL returned something broken
-        if resp.status_code != HTTPStatus.OK:
+        # URL returned something broken from the client side 4xx
+        # server side errors 5xx can pass for a retry
+        if (
+            HTTPStatus.BAD_REQUEST
+            <= resp.status_code
```
vkresch, line 387: If the response status code is greater or equal to BAD_REQUEST (400) and less then INTERNAL_SERVER_ERROR (500) the raise the APIProvider Exception which won't trigger a retry in the while loop.

src/rpc/rpc_reward_api.py
```diff
@@ -376,8 +380,13 @@ def do_rpc_request(self, request, time_out=120):
...
+            < HTTPStatus.INTERNAL_SERVER_ERROR
+        ):
```
vkresch, line 389: Relevant code section regarding [this issue](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/511). 
If the response status code is greater or equal to BAD_REQUEST (400) and less then INTERNAL_SERVER_ERROR (500) the raise the APIProvider Exception which won't trigger a retry in the while loop.

ericlavoie: Anything 400-499 is an error we shouldn't return, I would say anything bellow 499

vkresch: I think it makes sense to raise the exception just for client side errors which are 400-499 everything else should be able to return so that the while loop continues.

Are you suggesting that a status code 3 can exists and we should cover that edge case?
image

ericlavoie: Any chances of getting 3xx, 1xx? If almost none, than your code is fine. I agree 5xx should retry.

ericlavoie (approved):

nicolasochem (approved):
Wow, there is some big test refactoring going on here :)

This PR refactors some of the tezos rpc provider code to add a function `__get_response`. This function has a built-in retry mechanism which will only retry in case of 500 errors, and throw an exception in case of 400 errors, which is in general sensible.

tzkt provider already had such a mechanism. The number of retries is now a global constant applicable to tzkt provider too.

The retry logic is not added to tzstats. I assume that the behavior is already as desired in tzstats (especially since you added a test for it).

Cursory review of the code looks good. I tried dry-run payouts and it went fine. It's hard to do an actual test because I can't get a 500 error on-demand.

This PR introduces integration tests for the rpc provider and tzstats in the same fashion that tzkt.

All these integration tests are self-contained and do not call any external API (that I can tell).

Block rpc tests are new for all three providers and cover their "block api". The RPC response to the level and cycle call is mocked up. These tests cover all the existing functions in the block api (I think?)

`test_rpc_block_api` has 4 tests:

```
tests/integration/test_rpc_block_api.py::test_get_revelation PASSED
tests/integration/test_rpc_block_api.py::test_get_current_cycle_and_level PASSED
tests/integration/test_rpc_block_api.py::test_get_delegatable_baker PASSED
tests/integration/test_rpc_block_api.py::test_get_delegatable_non_baker PASSED
```


`test_tzstats_block_api` adds 4 tests:

```
tests/integration/test_tzstats_block_api.py::test_get_revelation PASSED
tests/integration/test_tzstats_block_api.py::test_get_current_cycle_and_level PASSED
tests/integration/test_tzstats_block_api.py::test_get_delegatable_baker PASSED
tests/integration/test_tzstats_block_api.py::test_get_delegatable_non_baker PASSED
```
The `test_tzkt_block_api` test is added, here is the output:


```
tests/integration/test_tzkt_block_api.py::test_get_revelation PASSED
tests/integration/test_tzkt_block_api.py::test_get_current_cycle_and_level PASSED
tests/integration/test_tzkt_block_api.py::test_get_delegatable_baker PASSED
tests/integration/test_tzkt_block_api.py::test_get_delegatable_non_baker PASSED
```

Rewards tests have been extended to tzstats and node rpc.

`test_tzstats_reward_api` adds two:

```
tests/integration/test_tzstats_reward_api.py::test_get_rewards_for_cycle_map PASSED
tests/integration/test_tzstats_reward_api.py::test_tzstats_terminate_404 PASSED
```

`test_rpc_rewards_api` actually tests the feature added as part of this PR.

```
tests/integration/test_rpc_reward_api.py::test_get_rewards_for_cycle_map PASSED
tests/integration/test_rpc_reward_api.py::test_rpc_terminate_404 PASSED
tests/integration/test_rpc_reward_api.py::test_rpc_contract_storage_404 PASSED
tests/integration/test_rpc_reward_api.py::test_rpc_contract_storage_500 PASSED
```

`test_tzkt_rewards_api` are left largely alone. 

The tests for the actual feature being added (raise an exception for 4xx error) is added in the tzstats and rpc api tests, but not tzkt. Why not? Maybe add it there too?

Reward api tests do not mock up RPC calls directly: they mock up the internal representation of the results of these calls in csv files. Following the tzkt model, these files are populated if they do not exist, by actually calling the RPCs. When they exist, they are not called again.

While tzkt tests a large variety of scenarios, the new reward api tests introduced for tzstats and node rpc only cover one scenario of a pretty old cycle (100) of the stake now baker.

All of this should be merged, because it adds coverage, but the methodology above should be improved. I will write a ticket.

Thank you for fixing the actual issue, adding a test to it and refactoring the test suite in the process to fit your needs.

vkresch:
@nicolasochem thanks for the review I added tests also for tzkt and enable the retry for the `get_reward_split` function. Please take a look again.

nicolasochem (approved):
OK. It looks like your last changes are all in the test, except some constants and one extra retry in the tzkt provider. I didn't redo a dry-run but it should be fine.

It also looks like you moved some test functions around to test utils since they were used by several providers. OK.

PR #550, opened 2022-02-01
fix black

nicolasochem:
Linting is failing after black update, fixing.


vkresch (approved):
Thanks for the fix!

Work Effort: 0h

PR #552, opened 2022-02-03
fix: pip package installation to return False on error

### LINKED ISSUE #551: Bug: Does this ever return something else then true?!
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/1446f36d783df6721dd2e9d2f7ee32cdeac3b476/src/main.py#L14

jdsika:
resolves #551 

Effort=0.5h


vkresch (approved):
Work effort: 0.5h

PR #554, opened 2022-02-17
Created new trd-data folder structure, see: baking_dirs.py

### LINKED ISSUE #465: Snake case for arguments
I find strange to use the underscore for multi-words arguments.

e.g. I'd prefer using `--dry-run` than `--dry_run`.

Is it just me? Nothing important of course, but I wondered if there's a particular reason.

### LINKED ISSUE #498: dry-run writes a csv file in "done" folder while no payout was done
Dry-run should not write a "done" file since no payment was done. ~With defaults params, it appears to be the case.~

~It has been speculated on slack that when using a custom reports directory, dry runs write a report as if it was really paid out.~

Can anyone confirm or point to the code where this may be happening?

### LINKED ISSUE #553: Bug: Standard path on Windows
If I want to test TRD in simulation mode on a Windows 10 machine and I do not specify a configuration path I am getting the following error:

> 2022-02-10 13:42:41,351 - MainThread - ERROR - [Process Life Cycle completing With Failure] Error Details: Unable to find any '.yaml' configuration files inside configuration directory(U:\/pymnt/cfg)

The default on Linux is "~" but I guess I would change in generally in favor for the main directory of TRD "." to avoid compatibity issues with Windows and Linux? Or better something that does not break backwards compatibility. So the current user directory on Windows maybe as default? I think the difference with slash and backslash must be resolved as well. #498 is somehow related and I think one PR resolving both could be useful.

The other way to look at it is to explicitly let TRD fail when Windows is detected and just have linux+mac support. But our docs do not explicitly mention that Windows is not supported. What do you think @vkresch ?

jdsika:
resolves #498
resolves #553 
unified fix #465 

Feature:
New folder structure which is compatible with old report structure but resolves conflict with simulations. Logs are now in the base directory by default.
```
# ~/pymnt/
# └──cfg
#    └── tz1xx.yaml
# └──logs
#    └── verbose_backup
#    │   └── app_verbose.gz
#    └── app.log
#    └── app_verbose.log
# └──simulations
#    └── tz1xxx
#       ├── calculations
#       │   └── 449.csv
#       └── payments
#           ├── done
#           │   └── 449.csv
#           └── failed
#               └── 449.csv
# └──reports
#    └── tz1xxx
#       ├── calculations
#       │   └── 449.csv
#       └── payments
#           ├── done
#           │   └── 449.csv
#           └── failed
#               └── 449.csv
```

Breaking changes:
* deprecate --reports_base
* deprecate --config_dir
* unified --log-file to --log_file
* introduce --base_directory

Improvements:
* logs by default in base directory
* args parsing is now having a validation function
* removed dead code
* help sections now shows folder structure

jdsika, effort=12h
ericlavoie, effort=2h
vkresch, effort=0.5h

### DIFF
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -48,7 +48,7 @@ Options
     Public node base URL. Default is ``https://mainnet-tezos.giganode.io``. This argument will only be used in case the reward provider is set to ``prpc``. This node will only be used to query reward data and delegator list. It must be an ARCHIVE node.
 
 ``-b --base_directory <path>``
-    Directory for reports and configurations. Default value: ``../trd-data``.
+    Directory for reports and configurations. Default value: ``~/pymnt/``.
 
 ``-D --dry_run``
     Run without injecting payments. Suitable for testing. Does not require locking.
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -49,6 +49,12 @@ Options
 
 ``-b --base_directory <path>``
     Directory for reports, configuration and logs. Default value: ``~/pymnt``.
+    The directory contains the following folders:
+    
+    1. ~/pymnt/cfg
+    2. ~/pymnt/simulations
+    3. ~/pymnt/reports
+    4. ~/pymnt/logs
 
 ``-D --dry_run``
     Run without injecting payments. Suitable for testing. Does not require locking.
@@ -80,8 +86,8 @@ Options
 ``--syslog``
     Log to syslog. Useful in daemon mode.
 
-``--log-file <path>``
-    Log output file name and full path. Default value: ``~/pymnt/logs/app.log``.
+``--log_file <path>``
+    Application log output folder path and file name. By default the logs are placed into the --base_directory e.g.:: ``~/pymnt/logs/app.log``.
 
 Examples
 --------
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -48,7 +48,7 @@ Options
     Public node base URL. Default is ``https://mainnet-tezos.giganode.io``. This argument will only be used in case the reward provider is set to ``prpc``. This node will only be used to query reward data and delegator list. It must be an ARCHIVE node.
 
 ``-r --reports_base <path>``
-    Directory to create reports. Default value: ``../trd-data``.
+    Directory for reports and configurations. Default value: ``../trd-data``.
 
 ``-D --dry_run``
     Run without injecting payments. Suitable for testing. Does not require locking.
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -47,7 +47,7 @@ Options
 ``-Ap --node_addr_public <url>``
     Public node base URL. Default is ``https://mainnet-tezos.giganode.io``. This argument will only be used in case the reward provider is set to ``prpc``. This node will only be used to query reward data and delegator list. It must be an ARCHIVE node.
 
-``-r --reports_base <path>``
+``-b --base_directory <path>``
     Directory for reports and configurations. Default value: ``../trd-data``.
 
 ``-D --dry_run``
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -48,7 +48,7 @@ Options
     Public node base URL. Default is ``https://mainnet-tezos.giganode.io``. This argument will only be used in case the reward provider is set to ``prpc``. This node will only be used to query reward data and delegator list. It must be an ARCHIVE node.
 
 ``-b --base_directory <path>``
-    Directory for reports and configurations. Default value: ``~/pymnt``.
+    Directory for reports, configuration and logs. Default value: ``~/pymnt``.
 
 ``-D --dry_run``
     Run without injecting payments. Suitable for testing. Does not require locking.
@@ -81,7 +81,7 @@ Options
     Log to syslog. Useful in daemon mode.
 
 ``--log-file <path>``
-    Log output file.
+    Log output file name and full path. Default value: ``~/pymnt/logs/app.log``.
 
 Examples
 --------
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -48,10 +48,7 @@ Options
     Public node base URL. Default is ``https://mainnet-tezos.giganode.io``. This argument will only be used in case the reward provider is set to ``prpc``. This node will only be used to query reward data and delegator list. It must be an ARCHIVE node.
 
 ``-r --reports_base <path>``
-    Directory to create reports. Default value: ``~/pymnt/reports``.
-
-``-f --config_dir <path>``
-    Directory to find baking configuration. Default value: ``~/pymnt/cfg``.
+    Directory to create reports. Default value: ``../trd-data``.
 
 ``-D --dry_run``
     Run without injecting payments. Suitable for testing. Does not require locking.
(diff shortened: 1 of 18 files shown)


ericlavoie:
You are removing the f parameter, and this will require that existing users will need to populate those new directory.  This is quite a break from previous. How do you will an existing installation migrate to this new structure.

jdsika:
> You are removing the f parameter, and this will require that existing users will need to populate those new directory. This is quite a break from previous. How do you will an existing installation migrate to this new structure.

yes, the new structure does make this parameter obsolete. In addition this is a safety guard as the user realsizes that the parameter is gone when trying to use it. Migration should be manual. The issue here was that the "improper" use of individual reporting directories lead to an override of specific files. This should never happen. Can you help me with this PR?

ericlavoie:
> > You are removing the f parameter, and this will require that existing users will need to populate those new directory. This is quite a break from previous. How do you will an existing installation migrate to this new structure.
> 
> yes, the new structure does make this parameter obsolete. In addition this is a safety guard as the user realsizes that the parameter is gone when trying to use it. Migration should be manual. The issue here was that the "improper" use of individual reporting directories lead to an override of specific files. This should never happen. Can you help me with this PR?

I have a few issues with removing the end users option to detemine the location of the root folder or name.
First scenario I run testnets on 2 chains, and run 2 TRD instance for these 2 testnets.
Second in migrating the folders manually if a user forgets to copy the payment folder, could it not start a payment process since it detects none were done.
What is the case to control the folder root name of an installation?

src/configure.py
```diff
@@ -439,7 +438,7 @@ def main(args):
     )
 
     # 1. find where configuration is
-    config_dir = os.path.expanduser(args.config_dir)
+    config_dir = os.path.expanduser(os.path.join(args.reports_base + CONFIG_DIR))
```
ericlavoie, line 441: Should it not use BASE_DIR constant if there is no argument for report dir?

jdsika: no, no. It was actually over defined before. Look at launch_common.py line 209 and "--reports_base"

jdsika:
Look at my comment. With defining a clear folder structure only ONE parameter is needed anymore and you can have different instances of TRD running properly with "--reports_base" and it actually does NOT override each others simulations anymore!!!

ericlavoie:
> Look at my comment. With defining a clear folder structure only ONE parameter is needed anymore and you can have different instances of TRD running properly with "--reports_base" and it actually does NOT override each others simulations anymore!!!

Ok what I understand "**report base**" is now defining the base **data dir**, it deprecates config_dir (-f) param, and it changes **report_dir** from **data_dir/reports** to **data_dir**.
For example I currently have i.e. --reports_base /home/user/datadir/reports.
To diminish confusion, would it not make sense to create a new param **data_dir** and deprecate report_dir param. 
I hope I read it correctly.

jdsika:
The default for `--reports-base` is `~/pymnt/reports` and if I am not mistaken then you find a subfolder with the baker address in there and you find this specific subfolder in your specific folder `--reports_base /home/user/datadir/reports`

If you migrate now you would need to use `--reports_base /home/user/datadir` to have your payment records in the folder `/home/user/datadir/reports/TZADDRESS` and the simulations in here `/home/user/datadir/simulations/TZADDRESS`

The sanity check here is the following:
`--config_dir ~/pymnt/cfg` is the default for the configy. The config is NOW placed here (in your case) `/home/user/datadir/configurations` - if you pass the -f argument it failes and if you do not pass it will not find it due to the remaning from `cfg `to `configurations` and we need to create a user message that guied through the migration. "please place the config file in here and move the reports to here". 

-D is already used for dry runs....  

What do you say?

jdsika:
@ericlavoie can you test `pay_for.py` on a testnetwork please?

jdsika:
@vkresch can you tell me if the class `ReleaseOverrideAction` in `configure.py` is used anywhere? In my opinion this is checked in `launch_common.py` in line 68?

jdsika:
TODO:
* depricate -r
* introduce e.g. -b
* put log file into data-dir
* create strucutre if not already existent and give error that migration is needed

src/Constants.py
```diff
@@ -1,5 +1,11 @@
 from enum import Enum
 
+# Persistent data directories
+BASE_DIR = "../trd-data"
```
vkresch, line 4: Not sure why you want to go one folder back to define the base directory? Seems very hacky ^^

docs/run.rst
```diff
@@ -48,10 +48,7 @@ Options
     Public node base URL. Default is ``https://mainnet-tezos.giganode.io``. This argument will only be used in case the reward provider is set to ``prpc``. This node will only be used to query reward data and delegator list. It must be an ARCHIVE node.
 
 ``-r --reports_base <path>``
-    Directory to create reports. Default value: ``~/pymnt/reports``.
-
-``-f --config_dir <path>``
-    Directory to find baking configuration. Default value: ``~/pymnt/cfg``.
+    Directory for reports and configurations. Default value: ``../trd-data``.
```
vkresch, line 51: I am not sure why ../trd-data ?

src/Constants.py
jdsika: just not in the git folder but one above? Is it hacky? What's your idea?

jdsika:
@vkresch 
`check_fee_ini( )` and` check_migration_complete( )` should both be in the state machine after the bakingDirs have been initialized. or?

src/Constants.py
```diff
@@ -1,5 +1,11 @@
...
+CONFIG_DIR = "/configurations"
+SIMULATIONS_DIR = "/simulations"
```
vkresch, line 6: Please remove the `/`. When we make `os.path.join(args.reports_base, SIMULATIONS_DIR)` later on the function `os.path.join` thinks you give it two absolute paths from which it takes second one. The diskfull regression test tries to create a /simulations folder in the root for which it then throws a Permission Denied error.

src/Constants.py
vkresch: not sure why you are in the git folder in the first place? maybe could let the user decide where he wants the base directory or just use the repo as base.

docs/run.rst
ericlavoie: probably better to go with ~/trd-data, and expand user when using it.

jdsika: OK, maybe the MAIN problem was the simulations always being in one place and the easiest solution is the current solution of my latest commit? How to expand ~ properly on windows btw?

ericlavoie: You use expand_user, for reference: https://www.geeksforgeeks.org/python-os-path-expanduser-method/
It's also in email.py iirc if you want an example.

jdsika: I think we are using that function everywhere already. I read that it should work on Windows as well.

src/Constants.py
jdsika: I changed some default values and I use the ~ again. It turns out that we needed to play a litte with the different os.path functions to get it properly returning correct paths for windows. The way I do it now is not breaking backward compatibility IMO

ericlavoie (approved):
Using os.path.normpath converts to lowercase the path, is that an issue

ericlavoie (approved):
Reverting to this naming convention makes sense to me, no migrating needed.
Looks good.

jdsika:
@vkresch do you see if `--dry_run_no_consumers` is even properly used somewhere?!

ericlavoie:
Tested deprecation of parameters, both are caught.
the code finds the config file and all directories.

ericlavoie:
Logs folder gets created properly in the correct default folder.

ericlavoie (approved):
From the basic test I did, it works correctly, it finds the correct existing directory, or creates them correctly if they don't exist.

PR #556, opened 2022-02-22
Add possibility to use env variables for smtp_user and smtp_pass

Pea-skillz:
* Permit to use environment variables for SMTP_USER and SMTP_PASS

Can be usefull in a gitops environment to avoid to store confidentials values in yaml file.
Secrets can be used for these values in a k8s deployment.

Signed-off-by: Pierre-Emmanuel Andre <pierre-emmanuel.andre@skillz.io>

* **Analysis**: Avoid the storage of user and password for the email plugin

* **Solution**: Use environment variables to provide these values

* **Implementation**: Check that env variables are defined and use them

* **Performed tests**: Tried with yaml values, both env variables and just with one env variable.

* **Documentation**: Documentation has been updated

**Work effort**: 1h


utdrmac (approved):
LGTM

PR #558, opened 2022-02-24
initial hangzhou => ithaca

nicolasochem:
Ithaca work in progress.

We change the current testnet to ithacanet.

General TODOs:
- [ ] Bump version to v11
- [ ] Replace all occurrences of Hangzhou with Ithaca (especially in the docs)
- [ ] Check all URLs for updates: Constants.py
- [ ] Check network constants: Constants.py
- [ ] Check fee constants: batch_payer.py
- [ ] Check/adapt API tests

## tzstats
?

## Tzkt

I changed the code for parsing the constants (constant names changed) and I was able to make payout work with the recently released tzkt backend: https://ithacanet.tzkt.io/opJcfaeaKWDSAULWdP1Cj5aEJngti1pr2pMuagUBBPkEhe6DVPN

So, it looks like we are almost there with tzkt backend.

TODOs:
- [ ] deal with estimated reward calculation

## RPC 

We are starting the work with RPC endpoint. Using RPC as a backend is not practical on hangzhou since it requires querying rights which freeze the node. This should be much better on ithaca, so RPC is actually a viable option for paying rewards.

`roll_snapshot` has been renamed to `stake_snapshot` and the way to query the endpoint has changed: you need to "target" a block within the cycle for which you want the snapshot. So far, I have hardcoded "head" so it does not work well accross cycles.

Note that the default endpoints rpc.ithacanet.teztnets.xyz now runs v12.0 final so it supports the `stake_snapshot` endpoint (it does not work in the release candidates of v12) so you can run this PR against the native endpoint.

With this I am able to perform reward computations, but it crashes at a later point. I have not gotten to debugging this part yet.

TODOs:

- [ ] make the RPC dry-run pass with the current codebase
- [ ]  adjust the snapshot retrieving query to the relevant cycle instead of `head`
- [ ] manage the hangzhou => ithaca transition (see below)
- [ ] add tzkt and tzstats support

### Note on the transition

On mainnet, during the first ithaca cycles, the stake snapshots will not be taken from the last hangzhou cycles. Instead, it will be taken from the transition block. This means that we need to "special case" these cycles, and write temporary code for that. The same thing can be tested on ithacanet since it started on hangzhou for the first 2 cycles.

@vkresch @jdsika


nicolasochem:
replaced by a new branch since this branch is now protected and I can't push to it anymore

PR #562, opened 2022-03-03
fixed self.args to args

jdsika:


ericlavoie (approved):
This now works, tested it

Issue #574, opened 2022-03-24
Installing requirements causes and error with the latest tweepy 4.7.0


EMES77:
**Describe the bug**
I installed TRD from scratch because i had a very old version running.
When running requirements.txt i got the error `tweepy 4.7.0 has requirement requests<3,>=2.27.0, but you'll have requests 2.21.0 which is incompatible`
I solved this locally by changing the first row in requirements.txt from `requests>=2.20.0` to `requests>=2.27.0`. After that no error message occured and i could start TRD as usual.
I'd change it myself and make a pull request but i don't know how (yet)

Running on Raspberry Pi 64bit. No docker.

jdsika:
Look at the developer section I try a PR. Thank you for the report! A very good start 🙂

ericlavoie:
Good catch. thanks

EMES77:
I looked at the PR. I understand just little but I'll try my best to test.

ericlavoie:
I created the PR

PR #575, opened 2022-03-24
Adjust requirement.txt package pyyaml and request to new minimum

ericlavoie:
---
name: Adjust requirement.txt package pyyaml and request to new minimum

---
This PR resolves the issue #574 . The following steps were performed:

* **Solution**: Adjust the minimum requirement in requirements text to avoid version 2.21 for request, also updated minimum version of pyyaml since there was a security issue with package earlier than 5.3.1 

* **Performed tests**: Run TRD with these new versions,

**Work effort**: 1h

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 requests>=2.20.0
 base58>=1.0.2
-pyyaml>=4.2b1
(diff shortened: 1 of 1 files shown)


nicolasochem (approved):
thank you

PR #577, opened 2022-03-26
TzKT API: new reward split layout (tenderbake ready)

m-kus:
Hello all, this is a complementary PR for the work being done in the #558
I've updated the TzKT-specific implementation of the `RewardApi` interface in accordance with the breaking changes listed here https://github.com/baking-bad/tzkt/pull/86

Summary:
- `get_rewards_for_cycle_map` no longer uses deprecated fields
- double preendorsing denunciation rewards and losses are now taken into account

All the test are passing (except for the state graphs), I've temporarily changed the TzKT base URL to `staging.api.tzkt.io` so that one can run them locally (CI seems to be failing for some reason), but it should be reverted as soon as we roll out the new version to prod.

### DIFF
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -45,7 +45,7 @@ TZSTATS_PUBLIC_API_URL = {
 
 # TzKT
 TZKT_PUBLIC_API_URL = {
-    "MAINNET": "https://api.tzkt.io/v1",
+    "MAINNET": "https://staging.api.tzkt.io/v1",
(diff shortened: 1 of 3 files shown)


jdsika:
@vkresch state graphs issue. There is also one issue related to that and I saw that the paths for generating the graphs in pytests seems wrong. also pytest seems to generate stuff that is not part of the .gitignore

jdsika:
fix has been incorporated in #581 - thank you!

PR #581, opened 2022-03-30
Ithaca

nicolasochem:
Ithaca work in progress.

We change the current testnet to ithacanet.

General TODOs:
- [ ] Bump version to v11
- [ ] Replace all occurrences of Hangzhou with Ithaca (especially in the docs)
- [ ] Check all URLs for updates: Constants.py
- [ ] Check network constants: Constants.py
- [ ] Check fee constants: batch_payer.py
- [ ] Check/adapt API tests

## tzstats
?

## Tzkt

I changed the code for parsing the constants (constant names changed) and I was able to make payout work with the recently released tzkt backend: https://ithacanet.tzkt.io/opJcfaeaKWDSAULWdP1Cj5aEJngti1pr2pMuagUBBPkEhe6DVPN

So, it looks like we are almost there with tzkt backend.

TODOs:
- [ ] deal with estimated reward calculation

## RPC 

We are starting the work with RPC endpoint. Using RPC as a backend is not practical on hangzhou since it requires querying rights which freeze the node. This should be much better on ithaca, so RPC is actually a viable option for paying rewards.

`roll_snapshot` has been renamed to `stake_snapshot` and the way to query the endpoint has changed: you need to "target" a block within the cycle for which you want the snapshot. So far, I have hardcoded "head" so it does not work well accross cycles.

Note that the default endpoints rpc.ithacanet.teztnets.xyz now runs v12.0 final so it supports the `stake_snapshot` endpoint (it does not work in the release candidates of v12) so you can run this PR against the native endpoint.

With this I am able to perform reward computations, but it crashes at a later point. I have not gotten to debugging this part yet.

TODOs:

- [ ] make the RPC dry-run pass with the current codebase
- [ ]  adjust the snapshot retrieving query to the relevant cycle instead of `head`
- [ ] manage the hangzhou => ithaca transition (see below)
- [ ] add tzkt and tzstats support

### Note on the transition

On mainnet, during the first ithaca cycles, the stake snapshots will not be taken from the last hangzhou cycles. Instead, it will be taken from the transition block. This means that we need to "special case" these cycles, and write temporary code for that. The same thing can be tested on ithacanet since it started on hangzhou for the first 2 cycles.

@vkresch @jdsika


src/NetworkConfiguration.py
```diff
@@ -64,22 +64,29 @@ def get_network_config_from_public_node(network_name):
...
     network_config_map["BLOCKS_PER_CYCLE"] = int(constants["blocks_per_cycle"])
-    network_config_map["BLOCKS_PER_ROLL_SNAPSHOT"] = int(
-        constants["blocks_per_roll_snapshot"]
+    network_config_map["BLOCKS_PER_STAKE_SNAPSHOT"] = int(
+        constants["blocks_per_stake_snapshot"]
     )
     network_config_map["BLOCK_REWARD"] = int(
-        int(constants["baking_reward_per_endorsement"][0])
-        * int(constants["endorsers_per_block"])
+        int(constants["baking_reward_fixed_portion"])
```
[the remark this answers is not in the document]
ericlavoie: Bonus is apart from endorsement reward, endorsement reward is the amount you can see when you pull protocol info on rewards.

nicolasochem:
Some progress here, I got the total reward amount to match between tzkt and RPC for cycle 75 for baker `tz1foXHgRzdYdaLgX6XhpZGxbBv42LZ6ubvE`.

There is still a mismatch in the delegate own balance during the snapshot:

With RPC:

```
| address                              | type            |    staked_balance | current_balance |  ratio | fee_ratio |        amount | fee_amount | fee_rate | overestimate | adjustm>
| ------------------------------------ | --------------- | ----------------- | --------------- | ------ | --------- | ------------- | ---------- | -------- | ------------ | ------->
| tz1foXHgRzdYdaLgX6XhpZGxbBv42LZ6ubvE | B               | 1,747,136,111,472 |               1 | 1.000… |    0.000… | 2,328,659,244 |          0 |     0.00 |        False |      Fa>
| tz1MYohN9ZP1qka9YHrVMe8LiJtqdkVpbeQJ | D               |   150,204,034,360 | 150,204,034,360 | 0.082… |    0.004… |   190,188,566 | 10,009,925 |     0.05 |        False |      Fa>
| tz1iyNM3Aff4JufmrtDsUFCchp8WCMU65bAh | D               |   100,136,022,567 | 100,136,022,567 | 0.054… |    0.003… |   126,792,377 |  6,673,283 |     0.05 |        False |      Fa>
| tz1WoM3j29FjeL1j6BdSLQntk2aenQtYTG1e | D               |   100,136,022,567 | 100,136,022,567 | 0.054… |    0.003… |   126,792,377 |  6,673,283 |     0.05 |        False |      Fa>
| OWNERS_PARENT                        | OWNERS_PARENT   | 1,396,660,031,978 |               0 | 0.799… |    0.000… | 1,861,529,433 |          0 |     0.00 |        False |      Fa>
| FOUNDERS_PARENT                      | FOUNDERS_PARENT |                 0 |               0 | 0.010… |    0.000… |    23,356,491 |          0 |     0.00 |        False |      Fa>
```

With tzkt:

```
| address                              | type            |    staked_balance | current_balance |  ratio | fee_ratio |        amount | fee_amount | fee_rate | overestimate | adjustm>
| ------------------------------------ | --------------- | ----------------- | --------------- | ------ | --------- | ------------- | ---------- | -------- | ------------ | ------->
| tz1foXHgRzdYdaLgX6XhpZGxbBv42LZ6ubvE | B               | 1,748,524,720,393 |               1 | 1.000… |    0.000… | 2,328,659,244 |          0 |     0.00 |        False |      Fa>
| tz1MYohN9ZP1qka9YHrVMe8LiJtqdkVpbeQJ | D               |   150,204,034,360 | 150,204,034,360 | 0.082… |    0.004… |   190,037,526 | 10,001,975 |     0.05 |        False |      Fa>
| tz1WoM3j29FjeL1j6BdSLQntk2aenQtYTG1e | D               |   100,136,022,567 | 100,136,022,567 | 0.054… |    0.003… |   126,691,684 |  6,667,983 |     0.05 |        False |      Fa>
| tz1iyNM3Aff4JufmrtDsUFCchp8WCMU65bAh | D               |   100,136,022,567 | 100,136,022,567 | 0.054… |    0.003… |   126,691,684 |  6,667,983 |     0.05 |        False |      Fa>
| OWNERS_PARENT                        | OWNERS_PARENT   | 1,398,048,640,899 |               0 | 0.800… |    0.000… | 1,861,900,409 |          0 |     0.00 |        False |      Fa>
| FOUNDERS_PARENT                      | FOUNDERS_PARENT |                 0 |               0 | 0.010… |    0.000… |    23,337,942 |          0 |     0.00 |        False |      F
```

See the delegate own balance is different between the two. Delegator balance is identical. I will get to the bottom of this tomorrow.

Also still TODO, estimated and ideal rewards. But, assuming that tzkt is right, tzkt actual is in good shape now.

It still needs a lot of testing, with different bakers and different backends.

nicolasochem:
as of the commit above, discrepancies between tzkt and RPC provider appear to have been resolved, both for `actual` and `ideal` payout mode. tzstats is still not recommended.

jdsika:
> as of the commit above, discrepancies between tzkt and RPC provider appear to have been resolved, both for `actual` and `ideal` payout mode. tzstats is still not recommended.

disable the usage of tzstats before its resolved. I am making a TODO

Twente-Mining:
imo its time to intergrate payouts into tezos binary's
to many times upgrades lead to problems

jdsika:
> imo its time to intergrate payouts into tezos binary's to many times upgrades lead to problems

Just use something else and write a letter. But great  that instead of contributing and just gaining from the work of others you come here to shit talk. Take your comments elsewhere if you don't contribute to a solution.

ericlavoie:
Yep like @jdsika said, you come in this code to shitpost, what is wrong with you. Payment system is nor core to tezos. Now help out out or gtfo

Twente-Mining:
wow such acidity

ericlavoie:
This is nit a forum, or twitter, this a work platform, about issues with code being proposed. GTFO of this site, find a forum or post an issue. How dense can you be.

nicolasochem:
I have not made much progress in the past 48 hours.

Current blocker is that I have not found the right formula to calculate the endorsement rewards for a future cycle (only needed in `-R -11` mode, pay estimated rewards) using the RPC provider. I am working on it with core devs.

I think we should not merge until running TRD with the same parameters with all 3 providers gives us the same calculations file bit-by-bit. Also, we used to have tests for it, it would be useful to have fresh tests as well, but they can probably come in in another PR.

If you want to help, you are welcome to test RPC and tzstats providers (you have to comment out [these lines](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/ithaca2/src/launch_common.py#L59-L65)) and report any discrepancies you can find. It's good to test pathological cases such as: double baking (not sure if it has happened yet), lost endorsement rewards, revelation rewards. Try ideal and actual rewards with `-R -5`.

As of the most recent commit I remain highly confident in this branch with tzkt provider, actual rewards, and `-R -5` (which is what most people use).

ericlavoie:
We could always make -R -11 disabled for now, and address it later.

nicolasochem:
I have re-enabled RPC.

I have verified that RPC and TZKT methods give me the exact same simulation csv file, in both `-R -5` and `-R -11`, for at least one baker. You are welcome to do more testing.

There are still a few discrepancies with tzstats being worked on.

But, we can probably clean up the tests and merge this soon.

There is special-case code in RPC for cycles 468-474 because the stake snapshot for 468-473 is hardcoded to the first ithaca block, and for 474 we use the last snapshot of the cycle 468 (RPC returns wrong info).

We have a bunch of pre-ithaca functional tests against mainnet that we need to throw away and redo. But I would rather wait to be out of this initial ithaca phase (cycles 468-474) before hunting for interesting mainnet test cases.

So, I think we can test a little bit, fix the failing tests, then merge this.

src/NetworkConfiguration.py
```diff
@@ -12,7 +12,7 @@
 def init_network_config(network_name, config_client_manager):
     network_config_map = {}
     node_addr = config_client_manager.get_node_url()
-    try:
+    if True:
```
chiptus, line 15: Do we need this if?

src/NetworkConfiguration.py
```diff
@@ -23,12 +23,12 @@ def init_network_config(network_name, config_client_manager):
...
-            "Failed to get network configuration constants from a local node ({}).".format(
-                node_addr
-            )
-        )
+    # except Exception:
+    #    logger.debug(
+    #        "Failed to get network configuration constants from a local node ({}).".format(
+    #            node_addr
+    #        )
+    #    )
```
chiptus, line 32: Remove commented code?

src/main.py
```diff
@@ -92,10 +114,16 @@ def start_application(args=None):
 
 if __name__ == "__main__":
     # Check the python version
-    if not sys.version_info.major >= 3 and sys.version_info.minor >= 6:
+    if (
```
baketzforme, line 117: Again, this isn't sitting right with me. I think the `not` should perhaps be pulled outside of the parentheses?

```
if not (
    X
    and Y
)
```

src/configure.py
```diff
@@ -504,10 +504,16 @@ def main(args):
 
 if __name__ == "__main__":
 
-    if not sys.version_info.major >= 3 and sys.version_info.minor >= 6:
+    if (
+        not sys.version_info.major >= PYTHON_MAJOR
+        and sys.version_info.minor >= PYTHON_MINOR
+    ):
```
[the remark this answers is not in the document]
baketzforme: This was marked as resolved even though nothing was changed. But the same code found in `src/main.py` (lines 117-120) was changed as per my suggestion. Perhaps I am missing some nuance to the difference between this code and the code in `src/main.py`, but I just wanted to flag this for a double check that this code is written as intended, because it still looks wrong to me.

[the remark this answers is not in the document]
nicolasochem: Fixed now, thank you.

jdsika:
@nicolasochem did you test with tzstats?

src/launch_common.py
```diff
@@ -91,8 +103,17 @@ def args_validation(args, argparser):
...
         release_override = args.release_override
-        if release_override not in [-11, -5, 0]:
-            argparser.error("release-override must be -11, -5 or 0. Default is 0.")
+        preserved_cycles = default_network_config_map[network]["NB_FREEZE_CYCLE"]
+        estimated_reward_override = -preserved_cycles * 2 - 1
+        frozen_reward_override = -preserved_cycles
+        if release_override not in [
+            estimated_reward_override,
+            frozen_reward_override,
+            0,
```
jdsika, line 112: Is 0 now deprecated? Shouldn't we exclude it here and make -5 as default?

nicolasochem: yes, 0 is now deprecated, and I agree -5 should be the default, but there are complications. I thought we could take the opportunity to migrate this parameter to the config file. We talked about this before, and you said we should have a generic way to pass parameter with either the config file or the CLI. I don't disagree, I think it requires big refactoring which is better done after this PR is merged. Let's leave the default to 0 now.

nicolasochem:
> @nicolasochem did you test with tzstats?

A little bit but I found discrepancies. It is currently disabled. I suggest we merge and fix it later.

jdsika (approved):

PR #582, opened 2022-03-30
Ithaca constant

Pea-skillz:
Not sure about the constants use here.
What do you think about this ?


src/Constants.py
```diff
@@ -50,22 +50,29 @@
...
+        # https://rpc.ithacanet.teztnets.xyz/chains/main/blocks/head/context/constants
         "NAME": CURRENT_TESTNET,
         "NB_FREEZE_CYCLE": 3,
         "MINIMAL_BLOCK_DELAY": 15,
         "BLOCKS_PER_CYCLE": 4096,
         "BLOCKS_PER_STAKE_SNAPSHOT": 256,
-        "BLOCK_REWARD": 20000000,
-        "ENDORSEMENT_REWARD": 78125,
+        "BLOCK_REWARD": 10000000,
+        "ENDORSEMENT_REWARDS": 10000000,
```
ericlavoie, line 77: That is incorrect, endorsement rewards for Ithaca is 1428
see https://api.ithacanet.tzkt.io/v1/protocols/current

src/Constants.py
```diff
@@ -50,22 +50,29 @@
...
         "NB_FREEZE_CYCLE": 5,
         "MINIMAL_BLOCK_DELAY": 30,
         "BLOCKS_PER_CYCLE": 8192,
         "BLOCKS_PER_STAKE_SNAPSHOT": 512,
+        # Fixed baking amount (10)+ bonus (10 in the best case)
         "BLOCK_REWARD": 20000000,
-        "ENDORSEMENT_REWARD": 78125,
+        # endorsing_reward = (1 - baking_reward_ratio) * (1 - bonus_ratio) * total_rewards
+        # = (1-1/4)*(1-1/3)*40
+        "ENDORSEMENT_REWARDS": 20000000,
```
ericlavoie, line 66: Endorsement reward is 78125 for hangzhou

jdsika: It is about Ithaca here in this branch

src/Constants.py
denver-s: `[5000000,2143]` should mean:

- 5000000 mutez = 5 tez is the `BAKING_REWARD_FIXED_PORTION`

- 2143 mutez = 0.002143 tez is the `BAKING_REWARD_BONUS_PER_SLOT`

denver-s: For example, this block https://ithacanet.tzkt.io/320093/ has 6561 endorsements and has received 5 tez as `BAKING_REWARD_FIXED_PORTION` and 4.058842 tez as `BAKING_REWARD_BONUS`.

We can calculate `BAKING_REWARD_BONUS` as such: 

`(6561 - (7000 - 2333)) * 2143` where `(7000 - 2333)` is the `CONSENSUS_THRESHOLD` so we give bonus only to the endorsement slots above that.

jdsika: so it is a max( ) function that is either 0 or positive?

denver-s: I think it would be correct to distinguish the constants as such:

```
"CONSENSUS_COMMITTEE_SIZE": 7000,
"CONSENSUS_THRESHOLD": 4667,
"BAKING_REWARD_FIXED_PORTION": 5000000,
"BAKING_REWARD_BONUS_PER_SLOT": 2143,
"ENDORSING_REWARD_PER_SLOT": 1428,
```

Same logic for mainnet. We can omit `"BAKING_REWARD_BONUS": 4999619`. Then when calculating rewards, we should call a function that calculates `BAKING_REWARD_BONUS = (BLOCK_ENDORSEMENTS -CONSENSUS_THRESHOLD) * BAKING_REWARD_BONUS_PER_SLOT`

jdsika: agree, I think, This is both valid for mainnet and testnet, right?

jdsika: ok, I pushed a commit. Something in that direction?

denver-s: Yes, but for mainnet it should be:

```
"BAKING_REWARD_FIXED_PORTION": 10000000,
"BAKING_REWARD_BONUS_PER_SLOT": 4286,
"ENDORSING_REWARD_PER_SLOT": 2857,
```

jdsika: Are you in slack? maybe do a quick call?

jdsika:
I think there is still a "freeze time" [look](https://tezos.gitlab.io/ithaca/consensus.html#rewards):

> Delegates can set an upper limit to their frozen deposits with the commands tezos-client set deposit limit for <delegate> to <deposit_limit>, and unset this limit with the command tezos-client unset deposit limit for <delegate>. These commands are implemented by using a new manager operation Set_deposits_limit. When emitting such a command in cycle c, it affects the active stake for cycles starting with c + PRESERVED_CYCLES + 1; the new active stake is taken into account when computing the frozen deposit for cycle c+1 already, however the user may see an update to its frozen deposit at cycle c + PRESERVED_CYCLES + MAX_SLASHING_PERIOD at the latest (because up to that cycle the frozen deposit also depends on the active stake at cycles before cycle c+1).

jdsika (approved):

Issue #600, opened 2022-06-02
Reminder: Ithaca2 branch


jdsika:
Some things that need to be done after the merge of #581 
* Debug and activate tzstats API
   * New version is [here](https://tzstats.com/docs/api#income-table) and seems to satisfy our needs
* Remove release_override and introduce early_payouts: True/False
* Introduce and fix tests again (cache)
* Rework network constants and remove unnecessary ones

Issue #547, opened 2022-01-17
integration tests should mockup actual API calls instead of internal model representation in csv


nicolasochem:
With #545, we are extending the current model of testing rewards calculation to all providers - previously tzkt was doing it this way.

Reward api tests do not mock up RPC calls directly: they mock up the internal representation of the results of these calls in csv files. Following the tzkt model, these files are populated if they do not exist, by actually calling the RPCs. When they exist, they are not called again.

This is confusing because it's undocumented, and it's also not the best we can do.

It would be better to mock the rpc calls at lower level: add an option to pytest to call the apis "for real" and record the response. Then save all the data returned from these api calls in source control. In normal mode, replay all these recorded api calls, do not call any external provider.

This will allow us to remove all these csv files from the integration test folder. Insted, we will save the raw data returned by the RPC.

There are several tools in python that can do that.

This would apply to all three providers.

As a bonus, we will have a simple way to figure out whenever any of the upstream APIs has changed: just rerun the tests agains "real" api endpoints, and compared with what we have saved: should be identical. Otherwise, the upstream apis have changed and we need to adjust the code.

This will also allow us to update this saved data more often: right now, the cycles we are using for our testing are very old, and we will have to change the tests to use new cycles after ithaca anyway.

Speaking of ithaca, it will also be useful to develop ithaca-compatible TRD by quickly generating mocked ithacanet data.

If this works, we can extend this to all the tests, not just integration: the entire test suite will be self-contained.

PR #610, opened 2022-08-19
Ithaca increase test coverage and fix tzstats

vkresch:
Changelog:
- Test reactivation
- API consistency test so that tzstats = rpc = tzkt
- Reward API cached data update
- Fix tzstats reward api
- Fix rpc reward api

Resolves https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/600

To fix tzstats I am currently assuming since tzkt api is working it is the standard to which we should compare values for now.

**Notes:** 
I readded some tests with actual API calls. I think the best way to currently test TRD is a mix between real and mock API calls.
With real API calls we can detect faster deprecated API endpoints IMO thus we can adjust/fix things quicker. The real API calls should only do non changing data calls. That means we call only data which is static and can never change e.g. the balance in the **past** cycles. I know it slows down the merge process and that there is a [discussion](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/547) about mocking all the API calls but that way we can make sure TRD works.

Mock API calls we should use if we want to test some recent data which will change in the future. E.g. the **current** balance of a delegator or the data is to big to query each time (performance issue).

**IMPORTANT:**
Please use in future PRs:
`@pytest.mark.skip(reason="no way of currently testing this")`
if you really need to disable a test instead of removing or commenting it out. 
Also provide a **reason** for that for future developers who do not know why the test was disabled.

**Issues detected during test activation (tick means this PR solved this issue)**:
- [x] Querying the baking rights via pRPC causes currently a 429 To many requests error. I created an issue at the tezos repo (https://gitlab.com/tezos/tezos/-/issues/3791). Let me know if this is the right place to post the issue. 
- [x] RPC reward API did not filter out zero staking balance delegates
- [x] Updated potential_endorsement_rewards of tzstats by divinding with 1e6 to approximatly match tzkt futureRewards

**Work effort**: 32h


nicolasochem:
Thanks, the suggestion to use a mix of mock and real rpc calls makes sense. Also, I'll make sure to skip the pytest tests instead of commenting them out.

tests/integration/test_tzkt_reward_api.py
```diff
@@ -80,18 +81,24 @@ def test_get_rewards_for_cycle_map(self, address, cycle):
         )
         tzkt_rewards = tzkt_impl.get_rewards_for_cycle_map(cycle, RewardsType.ACTUAL)
 
+        # TODO: Investigate why rpc staking balance is not equal to tzstats or tzkt
```
vkresch, line 84: Could there be a reason why the staking balance are not matching between rpc and tzkt/tzstats?

src/Constants.py
```diff
@@ -38,8 +38,8 @@
 
 # Public RPC
 PUBLIC_NODE_URL = {
-    "MAINNET": "https://rpc.tzkt.io/mainnet",
-    CURRENT_TESTNET: "https://rpc.tzkt.io/{}".format(CURRENT_TESTNET.lower()),
+    "MAINNET": "https://mainnet-tezos.giganode.io",
```
vkresch, line 41: Not sure why we replace the public rpc from https://mainnet-tezos.giganode.io to https://rpc.tzkt.io/mainnet. Any reason?

tests/integration/test_tzkt_reward_api.py
nicolasochem: Yes, could you check if there are still discrepancies? There should not be any. Last time I checked, TZKT and RPC agreed, but tzstats had a discrepancy.

If there are none, remove this comment

src/Constants.py
nicolasochem: More reliable I think? Giganode was rate limited and giving occasional errors.

vkresch: Ok. It seems that the tzkt rpc interface does not provide any information for historical cycles. Thus to keep the backwards compatibility I would for now prefer using giganode. Do you think we should maintain the correct reward calculations for historic cycles like before granada? if not we should drop the support and also remove the cycle tests. It might simplify the codebase.

tests/integration/test_tzkt_reward_api.py
vkresch: Currently the baking_rights are not queries correctly. I created an issue [here](https://gitlab.com/tezos/tezos/-/issues/3791). Is there  another way to get the baking_rights via rpc for a given historic cycle?

nicolasochem: I replied to the issue. Rights for a given cycle have to be queried at the right level. This level normally is calculated with `last_block_level_in_cycle`, so maybe there is a problem with this calculation. Can you give an example query where there is a discrepancy?

src/Constants.py
nicolasochem: No, we do not need to maintain correct calculations for pre-granada. Ideally we would replace them with post-granada test data. Using giganode is fine.

vkresch:
Need to implement some final tests for the reward api of tzstats and then the PR can be reviewed and merged.

src/tzstats/tzstats_reward_api.py
```diff
@@ -40,12 +41,17 @@ def get_rewards_for_cycle_map(self, cycle, rewards_type):
             self.blocks_per_cycle * self.consensus_committee_size
         )
         potential_endorsement_rewards = (
```
vkresch, line 43: @nicolasochem can you send me a resource regarding the calculation of the potential_endorsement_rewards? I am trying to match the tzkt futureRewards with this calculation. Currently they are not matching.

vkresch:
@nicolasochem @jdsika can you both test the the tzstats reward api? I would like to merge this PR as soon as possible. Thanks!

src/Constants.py
jdsika, line 41: Giganode is not available anymore!

vkresch: Since when? Or will it become deprecated in the near future?

jdsika: https://twitter.com/GigaNode/status/1570424065613271040?t=iy7EkkfjnPiUFUS3PbOsVA&s=19

src/tzstats/tzstats_reward_api.py
nicolasochem: @vkresch it's diligently :wink: documented on the tezos dev slack.

https://tezos-dev.slack.com/archives/CV5NX7F2L/p1649433246273169?thread_ts=1648854391.875409&cid=CV5NX7F2L

Drop me your email if you don't have access.

src/Constants.py
vkresch: Thanks for the heads up :D. What would be the alternative then? https://rpc.tzkt.io/mainnet ?

src/tzstats/tzstats_reward_api.py
vkresch: I don't have access :(
Mail: viktor.kreschenski@gmail.com

jdsika: I have invited you

src/Constants.py
jdsika: or https://mainnet.smartpy.io pick one you like and that works in your test

jdsika:
@vkresch I think you should have a look at your tests and then we can test it, right? I will be more active now and push a release

jdsika:
@vkresch please make sure this gets ready this week!

src/tzstats/tzstats_reward_api.py
vkresch: @nicolasochem I compared the tzkt and tzstats potentialRewards and they seem to not match. Am I missing something? Check[ this slack message](https://tezos-dev.slack.com/archives/CV5NX7F2L/p1670716905787589?thread_ts=1648854391.875409&cid=CV5NX7F2L) for more details.

vkresch: For now I added an approximation for the assert.

vkresch:
@jdsika @nicolasochem ready for review/merge.

src/tzstats/tzstats_reward_provider_helper.py
```diff
@@ -288,10 +276,17 @@ def get_cycle_total_stake(self, cycle):
...
         return staking_supply
 
-    def __fetch_current_balance(self, address_list):
+    def get_current_cycle(self):
+        uri = self.api + tip
+        resp = requests.get(uri, timeout=5)
+        root = resp.json()
+        return root["cycle"]
+
+    def fetch_current_balance(self, address_list):
```
rvermootenct, line 289: are we planning to use this externally? Am I missing where this is happening?

tests/integration/test_api_consistency.py
```diff
@@ -0,0 +1,193 @@
...
+def test_get_revelation(
+    address_block_api_tzkt, address_block_api_tzstats, address_block_api_rpc
+):
+    assert address_block_api_tzkt.get_revelation(
+        NORMAL_TEZOS_ADDRESS
+    ) == address_block_api_tzstats.get_revelation(NORMAL_TEZOS_ADDRESS)
+    assert address_block_api_tzkt.get_revelation(
+        NORMAL_TEZOS_ADDRESS
+    ) == address_block_api_rpc.get_revelation(NORMAL_TEZOS_ADDRESS)
+    assert address_block_api_tzstats.get_revelation(
```
rvermootenct, line 57: Is this a duplicate test as:
a = b
a = c
b = c

I.e if a == b and a == c, b must == c
And you're already checking that each revelation is working.

src/tzstats/tzstats_reward_provider_helper.py
vkresch: No I don't think so. I was just renaming it for better readability and later we can also unit test those functions since we can then access them externally. Currently the usage of the private prefixes is not used in the entire code base thus I would like to avoid using `__` for now.

tests/integration/test_api_consistency.py
vkresch: Just a redundant plausibility check :). You are right. I would still keep it just in case for some flakiness behavior.

tests/integration/test_phase.py
```diff
@@ -0,0 +1,227 @@
...
+PAYOUT_CYCLE = 500
+
+logger = logging.getLogger("unittesting")
+logger.setLevel(logging.DEBUG)
+logger.addHandler(logging.StreamHandler())
+
+
+class TestCalculatePhases(TestCase):
+
+    baking_config = (
```
rvermootenct, line 29: Could we use the make_config util to make this?

tests/integration/test_api_consistency.py
jdsika: They need to check the reaso in the logs... I bet on RPC timeouts

tests/integration/test_phase.py
vkresch: done

rvermootenct:
I have spent time looking through the code and the tests, as well as running and confirming that my reports are the same while on ghostnet and using dry run while on mainnet on the different reward network providers. From what I can see this PR is ready to go. Awesome work @vkresch 

I will add a disclaimer that I am still new to TRD and may have missed important questions.

rvermootenct (approved):
I have spent time looking through the code and the tests, as well as running and confirming that my reports are the same while on ghostnet and using dry run while on mainnet on the different reward network providers. From what I can see this PR is ready to go. Awesome work @vkresch

I will add a disclaimer that I am still new to TRD and may have missed important questions.

jdsika (approved):

Issue #612, opened 2022-08-30
Config option for setting minimum payment amount


Vlad1mir-D:
**Is your feature request related to a problem? Please describe.**
There is payment strategy implemented by multiple public bakers which excludes all payments below some specific threshold.

**Describe the solution you'd like**
It will be good to have a configuration option similar to `min_delegation_amt` which will allow to set a minimum payment amount (e.g. `min_payment_amt`) so all payments below specific threshold will be excluded. 

**Describe alternatives you've considered**
Currently I'm using a wrapper script which among other things performs simulation dry-run, then adjusts `min_delegation_amt` according to specific threshold and observed payment amounts discovered during simulation dry-run and finally performs payment run.

**Additional context**
Wrappers are good when it's not possible to modify source code of software which isn't the case so implementing this feature in TRD itself will allow to decrease wrapper complexity and also make possible to improve payment strategies implementations for all of the TRD users.

Issue #576, opened 2022-03-25
Payment Report CSV: missing items


jdsika:
**Describe the bug**
After doing e.g. a --dry_run of a specific cycle and look at the calculations csv I realized that certain entries are not part of the payment csv. I traced the issue back to the rules map regarding the redirect of payments. I think the type of transaction needs to be set to "merged" and a description needs to be added and the items need to be properly added to the payment csv.

jdsika:
Known error from analysis.
The payment_address in the calculations report is not working correctly as it should contain the address the value is mapped to. In general the logic seems to be working but it not reported correctly.

kalouo:
Hello. I have configured a redirect from a `KT1` to `tz1`. 

To add to @jdsika: while my payment was correctly redirected to the the `tz1`, the calculations file is _not_ showing the `tz1` as the `payment_address`.

PR #613, opened 2022-08-30
Add `min_payment_amt` configuration option to allow setting minimum payment amount (#612)

Vlad1mir-D:
Feature: Add `min_payment_amt` configuration option to allow setting minimum payment amount (#612)

---
name: Pull Request
labels: [enhancement](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/labels/enhancement)

---
IMPORTANT NOTICE:

This PR resolves the issue 612. The following steps were performed:

* **Solution**: Added support for `min_payment_amt` configuration option similar to [`payment_requirements.minimum_amount`](https://github.com/kalouo/breadcrumbs/wiki/2.-Configuration#payment_requirements) option which is implemented in the [Breadcrumbs](https://github.com/kalouo/breadcrumbs).

* **Implementation**: At first I did [attempt](https://github.com/Vlad1mir-D/tezos-reward-distributor/commit/1c83073573391cbbaf9af909bff2f8587a5111b3) to implement this option in [`calc/calculate_phase_final.py`](https://github.com/Vlad1mir-D/tezos-reward-distributor/commit/1c83073573391cbbaf9af909bff2f8587a5111b3#diff-5a212ccad57abfed4cbe7db9c1df13af42eff52ee224a3728e978d3f8ff53afd) but went nowhere. Then I realized it will be much easier to make a recursive call for [`calculate()`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/0b667e179447e31f47eb72edf41773057c7d1a08#diff-4bd82803d5e7462d2f2f3d9d1e120aa6e6e6ef988bfc556e91ac7b15b1ec2820R155) in [`calc/phased_payment_calculator.py`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/0b667e179447e31f47eb72edf41773057c7d1a08#diff-4bd82803d5e7462d2f2f3d9d1e120aa6e6e6ef988bfc556e91ac7b15b1ec2820) and this seems to be the right choice in my opinion.

* **Performed tests**:
  * Executed simulation runs for cycles 1-518 and compared results with calculations made without this feature being enabled
  * Using in production starting from cycle 500

* **Check list**:

  - [X] I extended the Sphinx documentation (if needed).

**Work effort**: 12h

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -126,6 +126,13 @@ Available configuration parameters are:
 
     min_delegation_amt : 10
   
+**min_payment_amt**
+  A minimum payment amount can be set here. If this value is set to 10, 10 tez are required as minimum. Inherits behavior of excluded delegates set for *min_delegation_amt*.
+
+  Example::
+
+    min_payment_amt : 10
+
 **reactivate_zeroed**
   True/False - If True, an account to be paid found with a 0 balance will be reactivated, incurring the necessary burn fee and storage, and rewards will be sent. If False, any account with a 0 balance will be skipped payment. This will be noted in the CSV report.
 
--- a/src/calc/phased_payment_calculator.py
+++ b/src/calc/phased_payment_calculator.py
@@ -35,6 +35,7 @@ class PhasedPaymentCalculator:
         owners_map,
         service_fee_calculator,
         min_delegation_amount,
+        min_payment_amount,
         rules_model,
     ):
         self.rules_model = rules_model
@@ -42,6 +43,7 @@ class PhasedPaymentCalculator:
         self.founders_map = founders_map
         self.fee_calc = service_fee_calculator
         self.min_delegation_amnt = min_delegation_amount
+        self.min_payment_amnt = min_payment_amount
 
     #
     # calculation details
@@ -51,15 +53,18 @@ class PhasedPaymentCalculator:
     # owners reward = owners payment = total reward - delegators reward
     # founders reward = delegators fee = total reward - delegators reward
     ####
-    def calculate(self, reward_provider_model, adjustments=None):
+    def calculate(self, reward_provider_model, adjustments=None, rerun=False):
 
         phase0 = CalculatePhase0(reward_provider_model)
         rwrd_logs = phase0.calculate()
 
         total_rwrd_amnt = int(reward_provider_model.computed_reward_amount)
-        logger.info(
-            "Total rewards before processing is {:<,d} mutez.".format(total_rwrd_amnt)
-        )
+        if not rerun:
+            logger.info(
+                "Total rewards before processing is {:<,d} mutez.".format(
+                    total_rwrd_amnt
+                )
+            )
         if total_rwrd_amnt == 0:
             logger.debug("NO REWARDS to process!")
             return [], 0
@@ -126,6 +131,32 @@ class PhasedPaymentCalculator:
         rwrd_logs, total_rwrd_amnt = phase_last.calculate(
             rwrd_logs, total_rwrd_amnt, adjustments
         )
+
+        # run phases again if calculated minimal payment amount is greater than configured
+        if self.min_payment_amnt and self.min_payment_amnt > int(
+            min([rl.adjusted_amount for rl in rwrd_logs if not rl.skipped])
+        ):
+            self.min_delegation_amnt = int(
+                min(
+                    [
+                        rl.staking_balance
+                        for rl in rwrd_logs
+                        if not rl.skipped
+                        and rl.adjusted_amount > self.min_payment_amnt
+                        and rl.staking_balance
+                    ]
+                )
+            )
+            logger.info(
+                "Setting min_delegation_amt to {:<,d} mutez due to min_payment_amt set to {:<,d}. Running calculations again.".format(
+                    self.min_delegation_amnt, self.min_payment_amnt
+                )
+            )
+            rwrd_logs, total_rwrd_amnt = self.calculate(
+                reward_provider_model, adjustments, True
+            )
+        elif rerun:
+            return rwrd_logs, total_rwrd_amnt
 
         # sort rewards according to type and balance
         rwrd_logs.sort(key=functools.cmp_to_key(cmp_by_type_balance))
--- a/src/config/yaml_baking_conf_parser.py
+++ b/src/config/yaml_baking_conf_parser.py
@@ -31,6 +31,7 @@ from model.baking_conf import (
     DEXTER,
     CONTRACTS_SET,
     REWARDS_TYPE,
+    MIN_PAYMENT_AMT,
 )
 from util.address_validator import AddressValidator
 from util.fee_validator import FeeValidator
@@ -72,6 +73,7 @@ class BakingYamlConfParser(YamlConfParser):
         self.validate_share_map(conf_obj, OWNERS_MAP)
         self.validate_service_fee(conf_obj)
         self.validate_min_delegation_amt(conf_obj)
+        self.validate_min_payment_amt(conf_obj)
         self.validate_address_set(conf_obj, SUPPORTERS_SET)
         self.validate_specials_map(conf_obj)
         self.validate_dest_map(conf_obj)
@@ -173,6 +175,18 @@ class BakingYamlConfParser(YamlConfParser):
             raise ConfigurationException(
                 "Invalid value:'{}'. {} parameter value must be an non negative integer".format(
                     conf_obj[MIN_DELEGATION_AMT], MIN_DELEGATION_AMT
+                )
+            )
+
+    def validate_min_payment_amt(self, conf_obj):
+        if MIN_PAYMENT_AMT not in conf_obj:
+            conf_obj[MIN_PAYMENT_AMT] = 0
+            return
+
+        if not self.validate_non_negative_int(conf_obj[MIN_PAYMENT_AMT]):
+            raise ConfigurationException(
+                "Invalid value:'{}'. {} parameter value must be an non negative integer".format(
+                    conf_obj[MIN_PAYMENT_AMT], MIN_PAYMENT_AMT
                 )
             )
 
--- a/src/configure.py
+++ b/src/configure.py
@@ -37,6 +37,7 @@ from model.baking_conf import (
     FOUNDERS_MAP,
     OWNERS_MAP,
     MIN_DELEGATION_AMT,
+    MIN_PAYMENT_AMT,
     RULES_MAP,
     MIN_DELEGATION_KEY,
     DELEGATOR_PAYS_XFER_FEE,
@@ -80,6 +81,7 @@ messages = {
     "supporters": "Add supporter address. Supporters do not pay bakery fee. Type enter to skip",
     "specials": "Add any addresses with a special fee in form of 'PKH,fee'. Type enter to skip",
     "noplugins": "No plugins are enabled by default. If you wish to use the email, twitter, or telegram plugins, please read the documentation and edit the configuration file manually.",
+    "minpayment": "Specify minimum payment amount in tez. Type enter for 0",
 }
 
 parser = None
@@ -195,9 +197,22 @@ def onmindelegation(input):
             input = "0"
         global parser
         parser.set(MIN_DELEGATION_AMT, float(input))
-        parser.validate_service_fee(parser.get_conf_obj())
+        parser.validate_min_delegation_amt(parser.get_conf_obj())
     except Exception:
-        printe("Invalid service fee: " + traceback.format_exc())
+        printe("Invalid minimum delegation amount: " + traceback.format_exc())
+        return
+    fsm.go()
+
+
+def onminpayment(input):
+    try:
+        if not input:
+            input = "0"
+        global parser
+        parser.set(MIN_PAYMENT_AMT, float(input))
+        parser.validate_min_payment_amt(parser.get_conf_obj())
+    except Exception:
+        printe("Invalid minimum payment amount: " + traceback.format_exc())
         return
     fsm.go()
 
@@ -395,6 +410,7 @@ callbacks = {
     "ownersmap": onownersmap,
     "mindelegation": onmindelegation,
     "mindelegationtarget": onmindelegationtarget,
+    "minpayment": onminpayment,
     "exclude": onexclude,
     "redirect": onredirect,
     "reactivatezeroed": onreactivatezeroed,
@@ -419,7 +435,8 @@ fsm = Fysom(
             {"name": "go", "src": "foundersmap", "dst": "ownersmap"},
             {"name": "go", "src": "ownersmap", "dst": "mindelegation"},
             {"name": "go", "src": "mindelegation", "dst": "mindelegationtarget"},
-            {"name": "go", "src": "mindelegationtarget", "dst": "exclude"},
+            {"name": "go", "src": "mindelegationtarget", "dst": "minpayment"},
+            {"name": "go", "src": "minpayment", "dst": "exclude"},
             {"name": "go", "src": "exclude", "dst": "redirect"},
             {"name": "go", "src": "redirect", "dst": "reactivatezeroed"},
             {"name": "go", "src": "reactivatezeroed", "dst": "delegatorpaysrafee"},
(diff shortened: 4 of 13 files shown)


jdsika:
It is planned for next week! I am still on vacation:))

Vlad1mir-D:
> It is planned for next week! I am still on vacation:))

Oh, didn't mean to bother you. Thanks for this reply!

src/configure.py
```diff
@@ -202,6 +204,19 @@ def onmindelegation(input):
     fsm.go()
 
 
+def onminpayment(input):
+    try:
+        if not input:
+            input = "0"
+        global parser
+        parser.set(MIN_PAYMENT_AMT, float(input))
+        parser.validate_service_fee(parser.get_conf_obj())
```
jdsika, line 213: Why validate service fee here? And is the text for the exception correct?

tests/smoke/test_launch_states.py
```diff
@@ -142,7 +143,7 @@ def test_wrong_api_base_url(args):
     MagicMock(return_value=""),
 )
 def test_wrong_node_end_point(args):
-    args.node_endpoint = "https://testnet-tezos.giganode.io:4433"
```
jdsika, line 145: I think the 4433 is on purpose as it should test a "wrong endpoint"?

src/configure.py
Vlad1mir-D: This is dumb copy-paste from `onmindelegation` which seems to be the same dumb copy-paste from `onservicefee`.
Fixed both `onmindelegation` and `onminpayment`.

Vlad1mir-D:
@jdsika Successfully resolved all notices, thank you for your time!

Vlad1mir-D:
@jdsika What do you think? Do I need to fix/improve any other parts?

Vlad1mir-D:
@jdsika I understand that TRD is an open-source software and there are no obligations exist for maintainers but I have a bunch of features developed for TRD and would like to donate them properly to the TRD community.
I'm not attempting to push you in any way but I want to know if there are some shortcuts I can use to speed-up the MR review process.
Thank you for your time!

docs/configuration.rst
```diff
@@ -126,6 +126,13 @@ Available configuration parameters are:
 
     min_delegation_amt : 10
   
+**min_payment_amt**
+  A minimum payment amount can be set here. If this value is set to 10, 10 tez are required as minimum. Inherits behavior of excluded delegates set for *min_delegation_amt*.
```
jdsika, line 130: I am having trouble understanding the "inherits behavior [...]" part. What exactly is inherited here? I am setting a value elsewhere and it is copied to min_payment_amt?

Btw I think the abbreviations like "amt" for "amount" are a curse in the code. In the config we are basically forced to use it due to consistency. I try to use proper names in the code where possible... this is more a general remark rather then a request to change anything as you are consistent with the rest of the namings here

jdsika:
I invited you to the org in order to create branches with "write rights"

src/config/yaml_baking_conf_parser.py
```diff
@@ -176,6 +178,18 @@ def validate_min_delegation_amt(self, conf_obj):
                 )
             )
 
+    def validate_min_payment_amt(self, conf_obj):
+        if MIN_PAYMENT_AMT not in conf_obj:
+            conf_obj[MIN_PAYMENT_AMT] = 0
+            return
+
+        if not self.validate_non_negative_int(conf_obj[MIN_PAYMENT_AMT]):
```
jdsika, line 186: It is the first time I become aware of the "validate_non_negative_int" function. That thing feels wrong... but not your cup of tea

src/configure.py
```diff
@@ -195,9 +197,22 @@ def onmindelegation(input):
             input = "0"
         global parser
         parser.set(MIN_DELEGATION_AMT, float(input))
-        parser.validate_service_fee(parser.get_conf_obj())
```
jdsika, line 198: Why are you deleting the check for "validate_service_fee"?

jdsika: Before it felt like a "copy paste" error but now the previous check was removed intentionally?

src/configure.py
```diff
@@ -419,7 +435,8 @@ def onprefinal(input):
             {"name": "go", "src": "foundersmap", "dst": "ownersmap"},
             {"name": "go", "src": "ownersmap", "dst": "mindelegation"},
             {"name": "go", "src": "mindelegation", "dst": "mindelegationtarget"},
-            {"name": "go", "src": "mindelegationtarget", "dst": "exclude"},
+            {"name": "go", "src": "mindelegationtarget", "dst": "minpayment"},
```
jdsika, line 438: Am I seeing the "inheritance" here you mentioned in the docs?

docs/configuration.rst
Vlad1mir-D: > I am having trouble understanding the "inherits behavior [...]" part. What exactly is inherited here? I am setting a value elsewhere and it is copied to min_payment_amt?

"inherits behavior" means there are no separate keyword like `mindelegation` available for "min payment threshold" in the `rules_map` allowing to set `TOF`/`TOB`/`TOE` specifically for `min_payment_amt` case so if payment excluded because of minimum payment threshold it will follow the rules set for `mindelegation`.

I foresee two options here:
* Put extended clarification into the documentation so it will be more or less obvious that `min_payment_amt` simply manipulates `min_delegation_amt` internally, follows all rules set for `mindelegation` in the `rules_map` and calculation won't reflect separate cause for payment excluded due to the `min_payment_amt` threshold.
* Implement separate `minpayment` key for the `rules_map` and add a separate exclusion reason which will properly reflect exclusion reason in the calculation report. Initially while taking another approach for `min_payment_amt` I did [implement all that stuff](https://github.com/Vlad1mir-D/tezos-reward-distributor/blob/minpayment-phased/src/config/yaml_baking_conf_parser.py#L404-L411) but then realized it could add unnecessary complexity to the configuration and decided to drop it in favor of making easier to review final implementation MR.

What do you think? Which is the better way to proceed?

src/configure.py
Vlad1mir-D: Yes, it's intentional as [`onmindelegation`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/2c15b08aa15b92769176643f0b27e8ea507655eb/src/configure.py#L194-L204) itself seems to be a copy-paste from [`onservicefee`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/2c15b08aa15b92769176643f0b27e8ea507655eb/src/configure.py#L139-L150).
Before this change [`onmindelegation`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/2c15b08aa15b92769176643f0b27e8ea507655eb/src/configure.py#L194-L204) was validated by [`validate_service_fee`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/2c15b08aa15b92769176643f0b27e8ea507655eb/src/config/yaml_baking_conf_parser.py#L163-L167) which is completely unrelated to `min_delegation_amt` in config but actually validates `service_fee` while proper validation [`validate_min_delegation_amt`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/2c15b08aa15b92769176643f0b27e8ea507655eb/src/config/yaml_baking_conf_parser.py#L169-L179) for [`onmindelegation`](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/2c15b08aa15b92769176643f0b27e8ea507655eb/src/configure.py#L194-L204) exists but not used at all.

I fixed this issue but if you like I could create a separate MR for this fix but I prefer not to as the size of this change is relatively small.

src/config/yaml_baking_conf_parser.py
Vlad1mir-D: Ok, I'll mark this as resolved in this case.

src/configure.py
Vlad1mir-D: No, this part of code simply (well, not so _simple_ as it _should_ be in my opinion) sets the order of options during the configuration stage.
I explained what I meant by "inheritance" [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/613#discussion_r974822521).

Vlad1mir-D:
> I invited you to the org in order to create branches with "write rights"

Thanks, I'll follow a basic procedure for all stuff I currently have in my TRD wrapper and try to move them into TRD one by one.

Vlad1mir-D:
> Btw I think the abbreviations like "amt" for "amount" are a curse in the code. In the config we are basically forced to use it due to consistency. I try to use proper names in the code where possible... this is more a general remark rather then a request to change anything as you are consistent with the rest of the namings here

Noted. I attempted to use the same abbreviations everywhere but will use full length variable names in further features.

src/configure.py
jdsika: No. perfectly fine. You just see that I am basically reading these parts of the code for the first time :D

docs/configuration.rst
jdsika: hm, I am not really a fan of implicit behavior - but the complexity is already high. I think I would opt for the minpayment key anyway and extend the docs for configuration for the user. It is also the thing that made me hesitate it he recursive loop that the min delegation is manipulated!
In addition the payment report is already incomplete (see #576 ) and I would like that fixed. Another puzzle with a double use of a keyword with implicit behaviour might not help there? What do you think?

jdsika: I understand correctly that otherwise the avoided payments would be marked with "min_delegation_amt" in the report? Do you think it is a false statement? Are there situations in which a min_payment_amt and a min_delegation_amt are not connected?
Thinking about it we could also couple the explanations of both configurations, write them under each other and explain the connection. In the morning I am leaning towards this :)

Vlad1mir-D: @jdsika 
> It is also the thing that made me hesitate it he recursive loop that the min delegation is manipulated!

I can attempt to fix https://github.com/Vlad1mir-D/tezos-reward-distributor/tree/minpayment-phased and implement this part without recursion but honestly I don't think there is any reason to feel any hesitation about recursion.

> In addition the payment report is already incomplete (see https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/576 ) and I would like that fixed.

I'm not using dry-run mode because currently it requires an access to the signer and this destroys the whole idea of "dry run" as I'm afraid to provide an access to signer. So I can't comment this issue.

> Another puzzle with a double use of a keyword with implicit behaviour might not help there? What do you think?

I'm not sure what you mean by "help" so I think it definitely won't help with this thing :)

> I understand correctly that otherwise the avoided payments would be marked with "min_delegation_amt" in the report?

In current implementation - yes, they will be marked with `min_delegation_amt`.

> Do you think it is a false statement?

I would say this could be a confusing statement but there many other parts in TRD that's really confusing to me and this is among others is the reason I decided not to add an additional keyword.

> Are there situations in which a min_payment_amt and a min_delegation_amt are not connected?

What do you mean by "connected"? There could be situations where both `min_payment_amt` and `min_delegation_amt` will be defined in config but only one of them in current implementation could "win" (i.e. exclude most of the payments as the exclusion starts from the bottom) so I don't see any issues except confusion such report could represent but it could be addressed in the documentation.

> Thinking about it we could also couple the explanations of both configurations, write them under each other and explain the connection. In the morning I am leaning towards this :)

Well, it's possible to keep both reasons in the report, i.e. when payment is excluded due to one of the `minpayment` *or* `mindelegation` - there will be only one notice and when payment excluded by both thresholds there will be two notices.


To summarize:
1. Add an additional keyword `minpayment` to the `rules_map` with the same options which currently available for `mindelegation`
2. Keep both `mindelegation` and `minpayment` exclusion reasons in the report whenever payment get excluded

Are these two things going to be enough for this MR to be included into the master branch?
Thanks!

src/calc/phased_payment_calculator.py
```diff
@@ -127,6 +132,32 @@ def calculate(self, reward_provider_model, adjustments=None):
             rwrd_logs, total_rwrd_amnt, adjustments
         )
 
+        # run phases again if calculated minimal payment amount is greater than configured
+        if self.min_payment_amnt and self.min_payment_amnt > int(
+            min([rl.adjusted_amount for rl in rwrd_logs if not rl.skipped])
```
vkresch, line 137: Why do we need to run the calculation again if the configured min payment is greater then the calculated one? I thought we then should not pay at all and just return right? Isn't that the intention of this feature? Please correct me if I'm misunderstanding the feature.

nicolasochem:
> I'm not using dry-run mode because currently it requires an access to the signer and this destroys the whole idea of "dry run" as I'm afraid to provide an access to signer. So I can't comment this issue.

Hi! A dry-run payout is a payout that has been calculated, signed, but not broadcasted. We exercise as much of the stack as possible (because it's useful) except we never inject the operation. There is no reason to be afraid (unless you are touching dry-run logic itself)

jdsika:
> > I'm not using dry-run mode because currently it requires an access to the signer and this destroys the whole idea of "dry run" as I'm afraid to provide an access to signer. So I can't comment this issue.
> 
> Hi! A dry-run payout is a payout that has been calculated, signed, but not broadcasted. We exercise as much of the stack as possible (because it's useful) except we never inject the operation. There is no reason to be afraid (unless you are touching dry-run logic itself)

#615 should have just removed that feature

src/calc/phased_payment_calculator.py
Vlad1mir-D: > Isn't that the intention of this feature? Please correct me if I'm misunderstanding the feature.

Intention of this feature is to exclude all payments the same way as `min_delegation_amt` so this code works as expected.

docs/configuration.rst
Vlad1mir-D: @jdsika What do you think?

Vlad1mir-D:
@jdsika @vkresch Hi! What's left for this MR?

Vlad1mir-D:
@vkresch It would be a pleasure to get a final list of notices that prevents this MR from merging until addressed as I'm really eager to donate all of changes I made in my own fork to the TRD community.

Vlad1mir-D:
@vkresch 

> It would be easier to review the code if we could see a regression test e.g. test_min_payment_amount.py in the regression folder which tests your feature.

Because there is no response for https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/613#discussion_r977091140 it would be better to get all notices at once as current implementation may change so do the tests and I prefer not to re-implement something multiple times if it's possible not to do so.

Vlad1mir-D:
`/unstuck` :smiley_cat:

src/calc/phased_payment_calculator.py
```diff
@@ -127,6 +132,32 @@ def calculate(self, reward_provider_model, adjustments=None):
...
+            logger.info(
+                "Setting min_delegation_amt to {:<,d} mutez due to min_payment_amt set to {:<,d}. Running calculations again.".format(
+                    self.min_delegation_amnt, self.min_payment_amnt
+                )
+            )
+            rwrd_logs, total_rwrd_amnt = self.calculate(
+                reward_provider_model, adjustments, True
+            )
+        elif rerun:
+            return rwrd_logs, total_rwrd_amnt
```
vkresch, line 159: Please remove the return statement and the rerun variable since it is redundant. The function calculate returns anyway with reward_logs and the total reward amount. Ofc it sorts and logs in between but it should be fine.

src/calc/phased_payment_calculator.py
```diff
@@ -51,15 +53,18 @@ def __init__(
     # owners reward = owners payment = total reward - delegators reward
     # founders reward = delegators fee = total reward - delegators reward
     ####
-    def calculate(self, reward_provider_model, adjustments=None):
+    def calculate(self, reward_provider_model, adjustments=None, rerun=False):
```
vkresch, line 56: rerun is not needed

vkresch (changes requested):
Thanks for the feature and sorry for the long review delay from my side. The implementation is good and the idea with the recalculation is a good example of code reusage :). The min payment is by default 0 if not defined in a config thus the feature is backwards compatible. I have just some minor change request and after that it can be merged. Thanks again and sorry for the inconvenience.

jdsika (approved):

Issue #614, opened 2022-09-05
Use TRD in dryrun without signer


redref:
**Description**
When using TRD in dryrun mode, signer configuration is still required to validate `payment address`.

**To Reproduce**
Start TRD in dryrun without signer started or configured with payment address key.

**Expected behavior**
Should compute payments and stats without halting.

**Environment**
 - Docker / K8S
 - testnet
 - `python src/main.py --network GHOSTNET --base_directory /var/run/tezos/trdconfig --reward_data_provider tzkt --signer_endpoint http://127.0.0.1:6732 --node_endpoint <endpoint> --docker --dry_run --payment_offset 20 --release_override 0`
 - Alpine 3.16

nicolasochem:
First and most importantly: thank you for this contribution.

I missed the train on this one but it's important to test signing when in dry-run mode.

Signing can go wrong for a variety of reasons, tezos has many different signer implementations, so it is really important for dry run to test the entire stack, which includes signing, and stop right before injection.

Arguably it's dangerous to sign operations because they can be replayed later. And the original issue is valid: when trying payouts with ledger for example, it's annoying to have to plug it, unlock it, just to try payouts.

Can we sign by default in dry-run and have a `--dry-run-no-signing` flag?

redref:
I tend to agree with you on the "dryrun should sign and drop before injection".

On the other end, modification was only about checking key exists in signer. Signature was and is not actually done in dryrun mode.

I am not sure I can handle modifications to do so.

PR #615, opened 2022-09-05
Do not check signer/validate address in dry run mode

redref:
---
name: Do not check signer/validate address in dry run mode

---

This PR resolves the issue #614 . The following steps were performed:

* **Analysis**:
During initialization, TRD checks key is present in signer but in my POV, should not use it in dryrun mode.

* **Solution**:
Forward dryrun flag to configuration loading and do not validate in this case.

* **Implementation**:
N/A

* **Performed tests**:
Running on my side to provide stats and payments CSVs (in another security context than the baker)

* **Documentation**: 
Please let me know if something needs to be done here.

* **Check list**:
Please let me know if something needs to be done here.

**Work effort**: 1h


redref:
Updated to fix linter.

jdsika:
I am trying to think hard if there could be any issue with this... but I think it is straight forward and you are right that this check is annoying :)

jdsika:
Please fill out your name [here](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/contributors.csv)

vkresch (approved):

Issue #621, opened 2022-09-25
uncaught payout failures


nicolasochem:
Follow up from #619 

In kathmandu, some rpc fields changed names resulting in a python exception when calculating payouts.

The exception was inproperly caught, resulting the failed payouts to NOT be listed in the `failed` folder. Instead, they were listed in the `done` folder with a status of `undefined`.

We are missing a try/catch somewhere, so when such errors happen, we mark payout as failed, and an updated TRD can then reprocess the payouts without needing to manipulate files by hand.

To replicate:

* revert https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/e1c5e4e668806f32b01eddeb02e8a29dff4c7943
* run a payout where some delegators are KT1

PR #623, opened 2022-09-26
Handle simulation errors and mark payments as failed

AndrewKishino:
---
name: Handle simulation errors and mark payments as failed
about: General try/catch for `simulate_single_operation` when calling `attempt_single_batch`
labels: 

---
IMPORTANT NOTICE:

This PR resolves the issue #621. The following steps were performed:

* **Analysis**: Some RPC fields changed names resulting in a python exception when simulating payments to KT accounts

* **Solution**: Wrap operation simulations in a try/catch and handle errors by marking the payment as failed

* **Implementation**: Catching general exceptions from the operation simulation and marking the payment as failed as well as logging the error.

* **Performed tests**: Reverted https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/e1c5e4e668806f32b01eddeb02e8a29dff4c7943 and ran payments for a cycle on Ghostnet. Verified failed payments to KT accounts are accurately recorded in `pymt/reports/<tz...>/payments/failed/<cycle>.csv`. Reset https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/e1c5e4e668806f32b01eddeb02e8a29dff4c7943 and re-ran TRD for the same cycle to process failed payments. Confirmed previously failed payments were successfully applied.

* **Check list**:

**Work effort**: 1hrs


jdsika:
We could argue if we put AVOIDED or FAILED in there.

AndrewKishino:
My understanding is that FAILED payments would be filtered into the `failed` reports and are retry-able whereas AVOIDED payments are not considered FAILED and are not retried.

jdsika:
Yes, line 707 is therefore inconsistent and probably wrong?

src/pay/batch_payer.py
```diff
@@ -687,9 +687,22 @@ def attempt_single_batch(self, payment_items, op_counter, dry_run=None):
...
+                    logger.info(
+                        "Payment to {} script could not be processed. Payment simulation failed with error: {}: {} ".format(
+                            payment_item.paymentaddress, type(e).__name__, str(e)
+                        )
+                    )
+                    payment_item.paid = PaymentStatus.FAIL
+                    payment_item.desc += "Payment simulation encountered an error while executing. Marking payment as failed. "
+                    continue
 
                 if simulation_status == PaymentStatus.FAIL:
```
AndrewKishino, line 707: `simulation_status` can be `FAILED` if `status != HTTPStatus.OK`. I agree with you that if the node/api is not reachable, that should not `AVOID` the payment, but mark it as failed so that it can be retried again when the node/api is available again.

AndrewKishino: If that is a fair assumption, then we could throw an exception when `status != HTTPStatus.OK` and it will be marked as failed instead of avoided.

jdsika: now there is still the mystery of double entires. As this catch was missing the exception was obviously cought one level above and it retried the whole batch. Maybe it is due to this?

jdsika: @AndrewKishino should we progress here? Do you want to finish it?

jdsika (approved):

Issue #491, opened 2021-08-14
FR: Remove producer/consumer; Confusing output


utdrmac:
There is no real purpose behind the P/C operation of TRD. It does not make TRD process any faster due to parallelism nor does it improve concurrency. What it does do is present confusion:

```
2021-08-14 13:19:25,125 - producer  - INFO - Calculation report is created at '/home/stuff/.trd/reports/tz1RpNqAzQNaZuxeLSbfEzVEzUWssddsL6Kw/calculations/383.csv'
2021-08-14 13:19:25,125 - producer  - INFO - Reward creation is done for cycle 383, created 1135 rewards.
2021-08-14 13:19:25,126 - producer  - INFO - Run mode ONETIME satisfied. Terminating...
2021-08-14 13:19:25,126 - producer  - DEBUG - Producer returning...
2021-08-14 13:19:25,735 - MainThread - INFO - Interrupted.
2021-08-14 13:19:25,735 - MainThread - INFO - TRD is shutting down...
2021-08-14 13:19:25,735 - MainThread - INFO - --------------------------------------------------------
2021-08-14 13:19:25,736 - MainThread - INFO - Sensitive operations are in progress!
2021-08-14 13:19:25,736 - MainThread - INFO - Please wait while the application is being shut down!
2021-08-14 13:19:25,736 - MainThread - INFO - --------------------------------------------------------
```

TRD remained like this for a full minute. Is TRD stuck? Looks like it is shutting down/stopping. But no, not really. In the background is another thread injecting the multiple batches of payouts.

IMO, we should remove the P/C in favor of a simple payout loop. The P/C is just unnecessary overhead/complexity.

utdrmac:
More on this. When using `-M 3` mode, the messages are confusing as it says TRD is shutting down, but not really. In fact, there are still many batches to process until it can shut down.

```
2022-04-09 19:04:10,231 - consumer0 - INFO - 1276 payments will be done in 5 batches
...
2022-04-09 19:04:14,189 - producer  - INFO - Reward creation is done for cycle 462, created 1571 rewards.
2022-04-09 19:04:14,189 - producer  - INFO - Run mode ONETIME satisfied. Terminating...
2022-04-09 19:04:14,189 - producer  - INFO - Sending KeyboardInterrupt signal.
2022-04-09 19:04:22,746 - MainThread - INFO - Interrupted.
[...]

PR #631, opened 2022-12-01
Refactor: update dry run options and cleanup validator 

rvermootenct:
---
name: Pull Request
labels: 

---
IMPORTANT NOTICE:

* **Analysis**: A user girl has many separate options for dry run. She is currently also forced into not using a signer.

* **Solution**: A user girl should have a a single dry run flag and not be forced into not using the signer.

* **Implementation**: I moved all of the dry run args into a singe dry run arg that has multiple choices, namely:
- signer (default)
- no_signer
- no_consumer
- no_signer_no_consumer

Allowing the user girl more granular control of how she would like to use the software. 

I also spent some time continuing to clean up launch common by moving all the validation stuff into it's own file, creating new functions for each validation step and tested them individually.  Before this was a bit of a black box and the nesting was getting very confusing. Tried to clean that up.

* **Performed tests**:  Wrote 14 new tests, ran through all dry mode options. I do not have the ability to run through a non-dry run mode at this time and would request that one is done before merge as it touches so much.

* **Documentation**:  I removed dry run no consumer from documentation. And added documentation for how dry run woks.
* **Check list**:

- [ x ] I extended the help section (if needed).

**Work effort**:  8

### DIFF
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -123,8 +123,6 @@ BUF_SIZE = 50
 DRY_RUN = {
     "SIGNER": "signer",
     "NO_SIGNER": "no_signer",
-    "NO_CONSUMER": "no_consumer",
-    "NO_SIGNER_NO_CONSUMER": "no_signer_no_consumer",
 }
 
 
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -121,9 +121,9 @@ DISK_LIMIT_SIZE = 5 * GIGA_BYTE
 BUF_SIZE = 50
 
 
-class DryRun(Enum):
-    SIGNER = 1
-    NO_SIGNER = 2
+class DryRun(str, Enum):
+    SIGNER = 'SIGNER'
+    NO_SIGNER = 'NO_SIGNER'
 
 
 class RunMode(Enum):
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -121,10 +121,10 @@ DISK_LIMIT_SIZE = 5 * GIGA_BYTE
 BUF_SIZE = 50
 
 DRY_RUN = {
-    'SIGNER': 'signer',
-    'NO_SIGNER': 'no_signer',
-    'NO_CONSUMER': 'no_consumer',
-    'NO_SIGNER_NO_CONSUMER': 'no_signer_no_consumer'
+    "SIGNER": "signer",
+    "NO_SIGNER": "no_signer",
+    "NO_CONSUMER": "no_consumer",
+    "NO_SIGNER_NO_CONSUMER": "no_signer_no_consumer",
 }
 
 
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -22,6 +22,7 @@ TESTNET_PREFIX = "ghost"
 TESTNET_SUFFIX = "net"
 CURRENT_TESTNET = (TESTNET_PREFIX + TESTNET_SUFFIX).upper()
 
+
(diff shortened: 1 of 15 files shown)


src/config/yaml_baking_conf_parser.py
```diff
@@ -207,7 +207,7 @@ def validate_payment_address(self, conf_obj):
             )
 
         if len(pymnt_addr) == PKH_LENGHT and pymnt_addr.startswith("tz"):
-            if not self.dry_run:
+            if not self.dry_run == DRY_RUN['NO_SIGNER'] or DRY_RUN['NO_SIGNER_NO_CONSUMER']:
```
rvermootenct, line 210: should swap this to just be if dry run == signer.

src/util/args_validator.py
```diff
@@ -0,0 +1,129 @@
...
+    network = network_validator(args)
+    payment_offset_validator(
+        parser, args, network['BLOCKS_PER_CYCLE'], network['NAME']
+    )
+    initial_cycle_validator(parser, args)
+    release_override_validator(
+        parser, args, default_network_config_map[args.network]["NB_FREEZE_CYCLE"], network['NAME']
+    )
+    dry_run_validator(args)
+    return base_directory_validator(args)
```
rvermootenct, line 129: just return args.

src/config/yaml_baking_conf_parser.py
rvermootenct: that is expected behaviour. 
In PR (615)[https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/615] It was assumed that we should never use signers in. 

I'm now saying if dry_run is not NO_SIGNER or NO_SIGNER_NO_CONSUMER please run check_pkh_known_by_signer

jdsika: I am sorry, in case you do -D and no argument it goes to default? What happens with the field if you don`t use -D at all? I think the confusing part is the implcit behavior of it is either seen and evaluated as a string which defines the type of dry_run or is not existent and therefore a no dry_run at all. I think the field should explicitly express the behavior?

jdsika:
Let's aim for a merge today

jdsika:
DID YOU TEST WITH A GHOSTNET BAKER?

jdsika:
No ENUM use?

src/Constants.py
```diff
@@ -119,6 +120,13 @@
 
 BUF_SIZE = 50
 
+DRY_RUN = {
+    "SIGNER": "signer",
```
jdsika, line 124: Better an ENUM here?

nicolasochem:
Thanks. What's a "Consumer"?

Maybe find a better description for this mode of dry-run. What does it not do?

As far as I understand, the "consumer" takes the theoretical payouts and puts them into discrete operations and batches.

What's the use case for running "dry-run no consumer"?

In my opinion, it would be much better to remove the concept of "producer" and "consumer" altogether, see [this discussion](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/491).

So at least let's not make it worse by surfacing this terminology to the user.

jdsika:
hm, you may be right. I mean we currently do expose this to the user. It would have been a refactor which explains the differerent exisitng types of dry-runs better. It was confusing to me at least. But we can just deprecate the "no_consumer" option and clean it up internally later?

jdsika (approved):

Issue #637, opened 2022-12-15
Dry run command not reading correctly


rvermootenct:
**Describe the bug**
When looking at the help options the choices are read as DryRun.NO_SIGNER instead of parsing what that constant is. Therefore a user thinks they need to write -D DryRun.NO_SIGNER instead of -D NO_SIGNER.

**To Reproduce**
1.  Run  python3 src/main.py --help
2. Read the instructions for dry run

**Expected behavior**
The choices should be "SIGNER" and "NO SIGNER"

**Additional context**
This issue was brought to attention in the TRD slack by user Leo Treasure

PR #638, opened 2022-12-15
fix: dry run choices should read correctly

rvermootenct:
---
name: Pull Request
labels: 

---
IMPORTANT NOTICE:

This PR resolves the issue [637](https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/637). The following steps were performed:

* **Analysis**: The help documentation was not correct.

* **Solution**:  Documentation should read SIGNER and NO SIGNER not DryRun.NO_SIGNER and DryRun.SIGNER.

* **Implementation**:  Used a string for the choices. 

* **Performed tests**:  Read the help section and ran a dry run. 

**Work effort**:  0.1


src/util/parser.py
```diff
@@ -176,8 +176,8 @@ def add_argument_dry(argparser):
         "2. no_signer: Do not use signer",
         action="store",
         choices=[
-            DryRun.SIGNER,
-            DryRun.NO_SIGNER,
+            "SIGNER",
```
jdsika, line 179: I must admit that I hate Python more with every day.. you are telling me the following:
1) the default case casts a type Bool
2) the default is casting an enum class:string
3) the choices are just strings

I think this is terrible. At least we should do DryRun.SIGNER.value ? and consequently the action should properly store it as a value of the class? Doesn't this bother you?

rvermootenct: It absolutely bothered me.
Thank you for the suggestion of the .value

src/util/parser.py
```diff
@@ -176,8 +176,8 @@ def add_argument_dry(argparser):
         "2. no_signer: Do not use signer",
```
jdsika, line 176: please parse string into docs from enum.value

jdsika (approved):

PR #652, opened 2023-01-24
fix: confirm all zero balance accounts are dealt with

rvermootenct:
---
name: Pull Request
labels: 

---
IMPORTANT NOTICE:

* **Analysis**: If an account needs to be reactivated there are cases where payment is attempted and gives false positives as the uninitialized account creates a rollback of payments.

* **Solution**: Confirm that all uninitialized payment accounts are handled correctly by either skipping or paying for initialized free.

* **Implementation**: Found in CalculatePhaseZero we are checking if the delegate.type == reward_log.TYPE_DELEGATOR and this issue can result in more than just a delegator. 

* **Performed tests**:  Ran a bunch of dry runs and ghostnet runs. Confirmed that reactivate_zerod works as expected and we don't see the rollback we are seeing before.  Also wrote a few unittests for CalculatePhaseZero

**Work effort**: 4h (took me a long time to recreate issue and then figure out why it was happening)

GIF TAX:


src/calc/calculate_phaseZeroBalance.py
```diff
@@ -29,11 +29,9 @@ def calculate(self, reward_logs, reactivate_zeroed):
 
             # KT1 accounts do not require reactivation on 0 balance
             if (
-                delegate.type == reward_log.TYPE_DELEGATOR
```
jdsika, line 32: because the redirection can mean that the target address is not of TYPE_DELEGATOR? What is this type used for? Why do you think this has been here in the first place as a condition?

rvermootenct: The question of the hour is why this was here in the first place. This issue arrised for me when I added a founder with no balance.

PR #661, opened 2023-03-27
adapt ghostnet constats + deprecate -R --release-override

jdsika:
* updated the ghostnet constants
* removed -R argument
* introduced --adjust_early_payouts argument
* still keeping release_override as variable in the code but reduced it to values 0 or -(preserved_cycles+1)
* fixed bug in parser which allowed for invalid arguments
* set sunset to 29.03.2023


tests/unit/test_args_validator.py
```diff
@@ -164,6 +166,7 @@ def test_validate():
     assert SUT == argparse.Namespace(
         initial_cycle=-1,
         run_mode=1,
+        release_override=0,
```
vkresch, line 169: For the test this needs to be readded since release_override does not gets deleted in the `build_parser` function.

jdsika:
now you are doing it my way @vkresch by just deleting the damn tests ;)))

vkresch:
@jdsika I deleted the rpc tests in the api consistency test due to the removal of rpc support of trd.

PR #663, opened 2023-03-28
fix mainnet constants for mumbai and 2 wrong ghostnet constants

nicolasochem:
These constants are necessary to calculate estimated rewards. To merge into @jdsika 's PR.

### DIFF
--- a/src/Constants.py
+++ b/src/Constants.py
@@ -76,20 +76,20 @@ DEFAULT_NETWORK_CONFIG_MAP = {
         # General
         "NAME": "MAINNET",
         "PRESERVED_CYCLES": 5,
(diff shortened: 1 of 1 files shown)


vkresch:
Thanks! Any source were we can double check these constants?

nicolasochem:
> Thanks! Any source were we can double check these constants?

Mumbai source code: https://gitlab.com/tezos/tezos/-/blob/master/src/proto_016_PtMumbai/lib_parameters/default_parameters.ml#L117-143

Or tomorrow you can simply query `chains/main/blocks/head/context/constants` on RPC after mumbai activates.

jdsika (approved):

PR #664, opened 2023-03-29
disable RPC after Mumbai

nicolasochem:
There are 2 problems:

* blocks per cycle is changing, so the calculation of cycle beginning, cycle snapshots etc... needs to be adjusted
* due to twice more blocks per cycle, the rights query will take even longer, making rpc provider even less appealing as an option

Besides, the engineering capacity is limited, and everyone is using TRD with tzkt. Let's disable RPC for now.

Effort: nicolasochem, 0.5h | jdsika, 0.5h


jdsika:
tzstats is working to my knowledge. Was it disabled before?

nicolasochem:
> tzstats is working to my knowledge. Was it disabled before?

I guess so. I found the code like this :)

jdsika:
Well. I just figured out that the whole thing is not working, great

jdsika:
I am fixing it

jdsika (approved):

vkresch:
There are currently tests which are stuck in an infinite loop on the CI but locally they are fine. I am currently disabling them. Will merge them shortly into master.

PR #668, opened 2023-03-30
fix adjusted_early_payout argument

nicolasochem:
Right now on master when I do `--adjusted_early_payouts`, it's taken into account.

but `--adjusted_early_payouts True` counts as false?!

This makes it take zero parameters, like `--docker` or `--retry_injected`.


jdsika:
I am actually confused by this report as it worked for me... I will check

jdsika (approved):

PR #685, opened 2023-11-01
Migrate from tzstats to tzpro

jdsika:
* **Check list**:

**Work effort**: 6h


Nikolay-everstake:
Yep it works for us. Thank you.

vkresch:
Any reasons for this migration? 
Can't we just extend the API with tzpro instead of replacing tzstats with it or is tzstat going to be deprecated?

jdsika:
It was deprecated and replaced. Here is the migration guide. Tzstats does not exist anymore:

https://blockwatch.cc/blog/migrating-to-tzpro/

vkresch:
Ok got it. Thanks for the context.

jdsika:
I tired to catch absolutely every occurance of tzstats and also renamed the folder to blockwatch (the company behind it) in order to pave the way to include new API Interfaces later (if needed)

src/blockwatch/tzpro_api_constants.py
```diff
@@ -0,0 +1,55 @@
...
+idx_cb_current_balance = 1
+idx_cb_delegator_address = 2
+
+
+def load_key_from_env_variables():
+    load_dotenv()
+    try:
+        key = os.getenv("TZPRO_API_KEY")
+    except:
+        verbose_logger.exception("Unable to load TZPRO_API_KEY from .env file!")
```
vkresch, line 50: How about we add the api key to the command line when using tzpro instead of forcing the user to create a .env file? We are currently not using envríronmental variables anywhere in the project.

vkresch: We might wanna setup then a test api key for our test pipeline. For that we might use the secrets manager of github actions.

vkresch:
@jdsika can I push commits on this PR?

src/blockwatch/tzpro_api_constants.py
ericlavoie: We have a config file, we put api for twitter and other in it, why not use it. Why spread the config.

jdsika:
> @jdsika can I push commits on this PR?

yes, please! I am on erics side to put it in the config. I did not think about (stupid)

src/blockwatch/tzpro_api_constants.py
vkresch: Even better :)

vkresch:
Not ready yet. I am fixing the tests now.

vkresch (approved):
There are still some failing tests due to API updates which needs further investigation and is out of scope for this PR. I also noticed that the enum comparisons are not working correctly and I am not sure why. E.g. `payment_item.paid == PaymentStatus.PAID` is `false` but `payment_item.paid.is_paid()` is `true`. That's really confusing. Best way to compare enums would be to compare their values for now:
`payment_item.paid.value == PaymentStatus.PAID.value`

Currently failing tests:
- test_attempt_single_batch_tz
- test_attempt_single_batch_KT
- test_batch_payer_total_payout_amount
- test_simulate_single_operation
- test_failed_simulate_single_operation
- test_reset
- test_archive

Effort: 8h

PR #686, opened 2023-11-04
Improvement: Requirements installation in venv

jdsika:
* moved startup checks to launch_common
* python venv mandatory due to Ubuntu > 23.xx
* cleaned up checks for better understanding

**Work effort**: 3h


jdsika:
@vkresch please a quick review here

nicolasochem:
Hi! Can we revert this? It no longer works in a container.

```
│ Checking python version ...                                                                                                                                                                │
│                                                                                                                                                                                            │
│ ... installed Python version 3.10 is greater then minimum required version 3.7. OK!                                                                                                        │
│                                                                                                                                                                                            │
│ Please make sure to activate a virtual environment for python due to breaking changes in Ubutu >= 23.XX:                                                                                   │
│ https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/                                                                                               │
│                                                                                               
```

When running containerized, there is little reason to use a venv... Can you make this check more specific? Like, check for the exact reason you are making this mandatory? What's the ubuntu 23 thing?

jdsika:
After Ubuntu 23 it is considered a bad practice to install python packages system wide as it may interfere with the system functionality. You cannot use pip install anymore without an error. I actually think it makes sense and haven't thought about it before. 

The message obviously does not tell the right story... 
https://fostips.com/pip-install-error-ubuntu-2304/?amp=1

nicolasochem:
yes, please remove the `in_venv()` check from main branch @jdsika

vkresch:
@jdsika I am removing the in_venv check in this PR https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/690. Let me know if it's fine. Imo the user is responsible for setting up his environment for TRD correctly and not TRD. We might actually also consider removing the requirements_installed check since we provide the requirements.txt file which list the packages the user should have installed before running TRD.

jdsika:
As long as the program shuts down gracefully if the packages are missing.

Issue #687, opened 2023-11-06
payment error. Cycle 669


SemElVik:
I'm using tzkt api and got this error.
_commit/c852b20ec30264c4cec1747696735f84edd42d48_

------------

```2023-11-06 10:24:05,423 - MainThread - INFO - Anonymous statistics will be collected. See docs/statistics.rst for more information.
2023-11-06 10:24:05,436 - MainThread - INFO - Loading baking configuration file /home/tz/pymnt/cfg/tz1ZCLE7.................... .yaml
2023-11-06 10:24:05,942 - MainThread - ERROR - [Process Life Cycle completing With Failure] Error Details: 'str' object has no attribute 'isEstimated'
2023-11-06 10:24:05,943 - MainThread - INFO - TRD is shutting down...
......

vkresch:
@SemElVik thanks for the report. It might be related to the latest commit. Please use commit https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/commit/d4fd27a8dce2d5de8a0454f8a7e76491ca3802c0 for now. I will look into it.

SemElVik:
he works. thx

SemElVik:
_/commit/100902ea9195764a23c3728def2319f83629dc06_


**python3 src/main.py -C 669 tzkt -M 3 -D**

```2023-11-07 09:57:50,953 - MainThread - INFO - --------------------------------------------
2023-11-07 09:57:50,953 - MainThread - INFO - BAKING ADDRESS is tz1Ldzz6k1BHdhuKvAtMRX7h5kJSMHESMHLC
2023-11-07 09:57:50,953 - MainThread - INFO - PAYMENT ADDRESS is tz1c9HEyZdAWXZCLE7h3KcZCyykjZr7VY3QM
2023-11-07 09:57:50,954 - MainThread - INFO - --------------------------------------------
2023-11-07 09:57:50,955 - MainThread - INFO - [Plugins] No plugins enabled
2023-11-07 09:57:50,956 - MainThread - ERROR - [Process Life Cycle completing With Failure] Error Details: Attribute tzpro_api_key not found in application configuration.
2023-11-07 09:57:50,957 - MainThread - INFO - TRD is shutting down...```

jdsika:
@SemElVik exactly what the error says :)

Add the parameter to the configuration. If you use tzkt you can leave an empty string and if you want to use tzpro you can make a free account (until 100k calls per month) at tzpro.io

PR #688, opened 2023-11-06
Port back direct comparison enum and fix API key

vkresch:
Resolves https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/687 and API key naming fix.

**Work effort**: 0.5h

### DIFF
--- a/src/config/yaml_baking_conf_parser.py
+++ b/src/config/yaml_baking_conf_parser.py
@@ -351,7 +351,7 @@ class BakingYamlConfParser(YamlConfParser):
                 "Defaults to 'actual' rewards payout type.".format(REWARDS_TYPE)
             )
 
-        if conf_obj[REWARDS_TYPE].isEstimated():
+        if conf_obj[REWARDS_TYPE] == RewardsType.ESTIMATED:
             raise ConfigurationException(
                 "Setting 'rewards_type' to 'estimated' is no longer supported.\n"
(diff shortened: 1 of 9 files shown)


src/config/yaml_baking_conf_parser.py
```diff
@@ -351,7 +351,7 @@ def validate_rewards_type(self, conf_obj):
                 "Defaults to 'actual' rewards payout type.".format(REWARDS_TYPE)
             )
 
-        if conf_obj[REWARDS_TYPE].isEstimated():
+        if conf_obj[REWARDS_TYPE] == RewardsType.ESTIMATED:
```
vkresch, line 354: This is strange that `conf_obj[REWARDS_TYPE]` can be a string which needs further investigations why this is happening. For now I just port back as a quick hotfix.

jdsika: Can't you test for a string and make a case handling here? Maybe there is no proper cast in die config loader?

jdsika (approved):

jdsika:
@vkresch can you please check for a consistent use of the variable name? tzpro_api_key or tz_pro_api_key

I think it should be this: tzpro_api_key

ericlavoie:
> @vkresch can you please check for a consistent use of the variable name? tzpro_api_key or tz_pro_api_key
> 
> I think it should be this: tzpro_api_key

I agree, it matches the domain and the brand.

vkresch:
sure I will update it

PR #689, opened 2023-11-19
Prepare-release/v12

jdsika:
- [x] Increment version
- [x] fix tzpro doc
- [x] Only mention .venv/ in the instructions
- [x] Check instructions
- [x] Update CI workflow

* Work effort: 2h


.env.example
```diff
@@ -0,0 +1 @@
+TZPRO_API_KEY = "XXXXXXXXXX"
```
vkresch, line 1: Is there a reason why you need to introduce this?

.gitignore
```diff
@@ -4,8 +4,7 @@
 */*/__pycache__/
 */__pycache__/
 docs/build/
-venv/
-env/
+.venv/
```
vkresch, line 7: Let's also ignore ./env and ./venv or are you specifially using the hidden folder name .venv for a virtual environment?

docs/testing.rst
```diff
@@ -1,5 +1,6 @@
 Tests
 ========
+In order to run test which rely on the TZPRO API you have to copy the `.env.example` file to `.env` and add your personal TZPRO API Key.
```
vkresch, line 3: for .env you would need to load it in with the function dot_env(). 
Just say that you need to set the environmental variable `export TZPRO_API_KEY="XXXXXXXXXX"` before running the tests. If not some tzpro dependent test might fail.

vkresch: For vscode you can put:

```
"python.testing.pytestArgs": [
    "--env",
    "TZPRO_API_KEY=XXXXXXXXXX",
    "tests"
],
```

in .vscode/settings.json

.gitignore
jdsika: yes, that is the recommendation in the official documentation and that is also what I took as error message instruction and what is in the readme

docs/testing.rst
jdsika: `TZPRO_API_KEY = os.environ.get("TZPRO_API_KEY")`

having this in the header of each test using tzpro seemed to be the simplest solution. We do it anyway in the CI pipeline with the GitHub secret. That's also the reason why I wanted to have it the same way in the local test environment. Needing it for the tests makes it somehow strange to have put it in the config. We would need a helper function in the utilities for tests that actually parses the whole config and hands out the key and is called for every test with tzpro. seems bad?

jdsika: Your example btw does not work for me

.gitignore
vkresch: Which official documentation or readme are you referring to?

jdsika: https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/

docs/testing.rst
jdsika: I am just not sure ehy I need your vscode setting?

docs/conf.py
```diff
@@ -49,17 +49,17 @@
 
 # General information about the project.
 project = "Tezos Reward Distributor (TRD)"
-copyright = "2021, see contributors.csv"
+copyright = "2023, see contributors.csv"
```
vkresch, line 52: 2024, happy new year :)

PR #691, opened 2024-01-11
Update twitter.py

ericlavoie:
Fixed the Twitter plugin, Twitter requires that we use the version 2 of their API. This fix changes to use the APIv2.

---
name: Pull Request: Fix/Twitter
labels: Fix

**Work effort**: 3h


src/plugins/twitter.py
```diff
@@ -25,6 +25,12 @@ class TwitterPlugin(plugins.Plugin):
...
         super().__init__("Twitter", cfg["twitter"])
 
+        # Must auth against Twitter API to get Token
+        auth = tweepy.OAuthHandler(self.api_key, self.api_secret)
+        auth.set_access_token(self.access_token, self.access_secret)
+
+        self.twitter = tweepy.API(auth)
+
         # Must auth against Twitter API to get Token
         auth = tweepy.OAuth1UserHandler(
```
jdsika, line 35: Is this here still needed? @ericlavoie

src/plugins/twitter.py
```diff
@@ -69,7 +76,13 @@ def send_payout_notification(self, cycle, payout_amount, nb_delegators):
         # Truncate message to max tweet length
         tweet = tweet[: self.MAX_TWEET_LEN]
 
-        resp = self.twitter.update_status(tweet)
+        # We use APIv2 to post, API 1.1 no longer allow posting tweet
+        client = tweepy.Client(
```
jdsika, line 80: line 32 says "self.twitter = tweepy.API" what is this self.twitter used for?

src/plugins/twitter.py
ericlavoie: I pushed an update

src/plugins/twitter.py
ericlavoie: it initialize tweepy API, self-twitter used on line bellow to get the displayname of the twitter account

jdsika (approved):

PR #679, opened 2023-08-22
fix #678 - Fix single-shot mode always exiting in error

nicolasochem:
We added code in #653 to have TRD send a non-zero exit status when anything wrong happens - including payout failures, or (relevant here) keyboard interrupts.

However, it turns out that when using TRD in runmode -2, -3, -4, the regular course of operations is to have the producer thread send a KeyboardInterrupt when done... which is then treated as error!

This is done with `_thread.interrupt_main()`. It's bad: humans send interrupts, programs should not. Instead, I'm suggesting that the producer use `SIGUSR1` to trigger a normal interruption.

This triggers the same code path as today, just with an exit status of SUCCESS. So I expect that the main thread will still wait for the consumer, and it won't actually exit before payouts are done, if there are any.

This allows kubernetes bakers (for example) to use a k8s cronjob and get alerted automatically in case of payment failures.

For the record, I still believe this entire producer/consumer logic should be thrown away and we should simply do things in order (calculate, then pay). I've ranted about this before (#491).

**Work effort**: 3 hours

PR #695, opened 2024-02-07
signals workaround for windows

nicolasochem:
SIGUSR1 and SIGUSR2 introduced in #679 don't work on windows. For windows, I revert to the previous behavior (always exit succesfully) by selectively importing signals based on OS. It works for me on linux, I don't have a windows machine to try.

Work: 1hr

### DIFF
--- a/src/pay/payment_producer.py
+++ b/src/pay/payment_producer.py
@@ -1,5 +1,6 @@
 import _thread
 import os
+import platform
 import signal
 import threading
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


jdsika:
Was there a reason you used _signals instead of signals ? This seems to work.

jdsika (approved):

nicolasochem:
> Was there a reason you used `_signal` instead of `signal` ? This seems to work.

just to make clear to the reader that it's our own internal variable and not some standard python var. but your commit is ok. good to merge as is.

Issue #616, opened 2022-09-05
Loop stuck when baker not active in payment cycle


redref:
**Describe the bug**
TRD tries to compute on `payment cycle`, but get stuck on this cycle if baker was not active (no delegators, no rewards) for this cycle.

**To Reproduce**
1. Bootstrap a new baker address
2. Start TRD
3. TRD should be stuck at payment cycle (current - 3 - X)

**Expected behavior**
Payment computations should proceed to next cycle in this case.

**Screenshots**
```
2022-09-02 13:31:46,773 - producer  - ERROR - [try_to_pay] API provider error Max sequent calls number exceeded (256)
2022-09-02 13:31:46,774 - producer  - ERROR - tzkt error at payment producer loop: 'Max sequent calls number exceeded (256)', will try again.
```

**Environment**
 - Docker
 - Both (tested in testnet)
 - `python src/main.py --network GHOSTNET --base_directory /var/run/tezos/trdconfig --reward_data_provider tzkt --signer_endpoint http://127.0.0.1:6732 --node_endpoint <endpoint> --docker --dry_run --payment_offset 20 --release_override 0`
 - Alpine 3.16

**Additional context**
I have pinpointed issue to the TZKT api response NO_CONTENT treatment.
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/tzkt/tzkt_api.py#L89
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/blob/master/src/tzkt/tzkt_api.py#L194

It seems TZKT return 204 in this case (no delegators, no rewards) for this cycle. This is processed as a request error, and looping indefinitely on this cycle as a result.

PR #707, opened 2024-07-02
fix: tzkt api may return more delegators than numDelegators

TPXP:
---
name: fix: tzkt api may return more delegators than numDelegators
labels: bugfix, tzkt_api

---

* **Analysis**: TZTK seems to return more delegators than `numDelegators` in some cases on the `rewards/split` endpoint. Handle this gracefully and move on. See `tz3LV9aGKHDnAZHCtC9SjNtTrKRu678FqSki` at cycle 751 for an example: https://api.tzkt.io/v1/rewards/split/tz3LV9aGKHDnAZHCtC9SjNtTrKRu678FqSki/751?offset=0&limit=10000

* **Solution**: Update the length condition to stop if we fetched at least as much delegators as `numDelegators`. Also, add another safeguard by stopping the loop if we're not fetching any more validators (indicating we've reached the end of the list) - this way we will stop if TzKT returns less validators than `numDelegators`.

* **Performed tests**: Run TRD for baker `tz3LV9aGKHDnAZHCtC9SjNtTrKRu678FqSki` for cycle 751, confirm it works

* **Documentation**: N/A

* **Check list**:

- [x] I tested my changes and confirm they fix the issue I was facing

**Work effort**: 2 hours

Possibly fixes https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/issues/616

### DIFF
--- a/src/tzkt/tzkt_api.py
+++ b/src/tzkt/tzkt_api.py
@@ -212,7 +212,7 @@ class TzKTApi:
                 assert isinstance(res, dict) and "delegators" in res
                 res["delegators"].extend(page["delegators"])
 
-            if not fetch_delegators or len(res["delegators"]) == res["numDelegators"]:
+            if not fetch_delegators or len(page["delegators"]) < limit:
                 return res
             else:
                 offset += limit


nicolasochem:
thanks @TPXP , can you clarify what happens when tzkt returns more delegators than numDelegators? Who are the extra delegators in this case? Are they duplicates? what's happening with the payouts? Is the program crashing?

TPXP:
Hello @nicolasochem , thanks for the reply.

When TzKT replies more delegators, the `len(res["delegators"]) == res["numDelegators"]` condition never triggers so we make more and more calls with increasing offset. TRD ends up failing to calculate rewards for the cycle with the following exception.

```
producer  - ERROR - tzkt error at payment producer loop: 'Max sequent calls number exceeded (256)', will try again.
```

I couldn't find any duplicates in the example I'm giving - maybe a few weird addresses but that's all. It seems that `numDelegators` is just wrong. Here's how I checked:
```bash
$ wget https://api.tzkt.io/v1/rewards/split/tz3LV9aGKHDnAZHCtC9SjNtTrKRu678FqSki/751\?offset\=0\&limit\=10000 -O rewards.json
$ jq '.numDelegators' ./yolo.json
3224
$ jq '.delegators[].address' ./yolo.json | sort | uniq | wc -l
3225
$ jq '.delegators[].address' ./yolo.json | sort | less
"KT1JBkpjozcKXWpHJLVhU67ZpXFhzH9Hekzg"
"tz1KeBQPzzGA7rBDcUY7biCKEKTSnxX291Hm"
...
"tz1iyXtwz3bKa6r8KEh6CnQU6CayU4msz4zj"
"tz2VJ6pJnnb9rYHFSQzUiZHX1M3uGnvGu4pz"
# Maybe 0 balances? (No)
$ jq '.delegators[].delegatedBalance' ./rewards.json | sort -h | less
154
[...]
6025854993128
```

nicolasochem:
@Groxan can you please check on your end why this would be happening?

Groxan:
That indeed may happen, when an account re-delegates to another baker, but still has unstake requests with the previous baker. I.e. the account has delegated balance with two different bakers at the same time, but counted as delegator only with the actual one.

Groxan:
The best way to check for end of data is `response.length < limit`.

TPXP:
Thanks for confirming @Groxan , I've updated the PR to fetch pages until we get less delegators than `limit`.

darrensteele (approved):
Looks good to me - works fine in my bakery

nicolasochem:
thanks everyone

PR #708, opened 2024-07-10
fix cont payouts: the proper num blocks per cycle & time bw blocks must be set

nicolasochem:
Previously I had added code to detect the first Paris level.... but continuous mode is still broken, because it's calculating the new cycle wrong, because of the wrong constants!

Also:

* remove unused constants
* removed adjusted_early_payout param

**Work effort**: 2h


nicolasochem:
Rich Ayotte confirmed it was working.

PR #711, opened 2024-07-25
refresh docs

nicolasochem:
remove some outdated references, make it clear that staking and delegating are different


vkresch (approved):
Thanks for the update!

PR #712, opened 2024-07-25
Remove owners

nicolasochem:
TRD has the concept of "owners" where several tez holders can pool their tez in one bakery. TRD would then distribute rewards amongst them. The problem is, since Paris upgrade, rewards are split in two, delegated and staking rewards. For a baker, it makes sense to stake most tez, to increase staking & delegation capability. But then, rewards are staked (frozen) and TRD can not distribute them.

We are therefore removing the concept of "bakery owners" since it conflicts with the concept of staker and can lead to false expectations. In fact, it confused quite a few bakers during paris rollout.

We are leaving the concept of "founders" since the goal is to distribute the delegation fees to different accounts. This still makes sense as delegation fees are unfrozen and can be distributed. I would have preferred to remove it as well: indeed, staking does not have this capability, so why would TRD have it? However, after few people mentioned on slack that it is useful to them, I am leaving it.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -81,24 +81,6 @@ Available configuration parameters are:
   
   Alice and Bob are both founders of Acme Bakery. Their bakery fee is 10%. Alice has a 60% (0.6) interest in the bakery, and Bob has a 40% (0.4) interest. When rewards are delivered at the end of each cycle, 10% is taken as the bakery fee (ie: *service_fee*). That 10% is then divided between Alice and Bob according to their ratios.
   
-**owners_map**
-  A dictionary of PKH and ratio ( decimal in the range [0-1]) pairs. Each item in this dictionary represents PKH of each balance owner and his ratio of the amount he owns in the total baking balance. Implicit or originated addresses are accepted. It is important that the sum of all ratios equals to 1. This map is optional if owners do not want to be paid for baking rewards, in this case, baking rewards remain in baking balance.
-  
-  Example::
-
-    Current Baker Balance: 17,400 tez
-    Total Delegations: 69,520 tez
-    Total Staked: 86,920 tez
-
-    service_fee: 9
-    owners_map:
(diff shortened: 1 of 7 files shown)


nicolasochem:
> Our concern is, with this PR, will payments like the above go away? If they do go away, then wouldn't delegated rewards for the unstaked balance in the baker be left unpaid?

@vchong yes, you are correct. Owners still get the delegation rewards for the unstaked balace of the baker.

My point is:

* bakers should really not have an unstaked balance above 1 tez to pay for operation fees. delegated balance earns less than staking.
* it is really confusing for "owners" that their delegated rewards are paid out by TRD and their staked rewards are paid manually by the bakery operator.

I understand that the baker may have unstaked balance because they want to hold on to liquid tez to potentially sell them at short notice. Still, since the operator is making calculations to distribute staking rewards for all the bakery stakeholders, they can as well do manual delegation reward distribution.

My goal is primarily to remove some code to make the project easier to maintain. But if you are telling me that "owners" feature is still useful, I have no problem keeping it and closing this PR.

ericlavoie:
> > Our concern is, with this PR, will payments like the above go away? If they do go away, then wouldn't delegated rewards for the unstaked balance in the baker be left unpaid?
> 
> @vchong yes, you are correct. Owners still get the delegation rewards for the unstaked balace of the baker.
> 
> My point is:
> 
> * bakers should really not have an unstaked balance above 1 tez to pay for operation fees. delegated balance earns less than staking.
> * it is really confusing for "owners" that their delegated rewards are paid out by TRD and their staked rewards are paid manually by the bakery operator.
> 
> I understand that the baker may have unstaked balance because they want to hold on to liquid tez to potentially sell them at short notice. Still, since the operator is making calculations to distribute staking rewards for all the bakery stakeholders, they can as well do manual delegation reward distribution.
> 
> My goal is primarily to remove some code to make the project easier to maintain. But if you are telling me that "owners" feature is still useful, I have no problem keeping it and closing this PR.

PR #713, opened 2024-10-01
exit main thread with same exit code as payment_producer

nicolasochem:
https://github.com/tezos-reward-distributor-organization/tezos-reward-distributor/pull/679 introduced support for exit codes, so an alert can be sent in single-shot mode when payouts fail for any reason.

However, it was crude, only supporting exit code 1.

The producer thread supports many exit codes.

In this case, there is a benign issue where tzkt returns "not synced" and therefore payouts fail, but this is likely temporary and will pass at next try, so there is no need to alert. But, currently it's not possible to behave differently based on the exit code because it's always 0 or 1.

An ugly solution is to save the exit code of the child thread in a file, then read it in the main thread. That's what I am doing here. I remain convinced that the entire thread architecture needs to go away, and we need to make TRD single threaded again, but that's for another day.

Also:
* change the exit code of misconfigured provider to GENERAL_ERROR because it's not really a provider error,
* change the help to remove old providers that we don't support anymore

---
name: Pull Request
labels: 

---
IMPORTANT NOTICE:

* **Analysis**: If the described issue is a bug report, analyze the reasons resulting in this bug.

* **Solution**: Describe the proposed solution for the bug or feature.

* **Performed tests**: Describe the performed tests.

* **Check list**:

**Work effort**: Give your estimate of the work effort in hours. This might be adjusted or discussed by the other contributors in order to keep a fair rewarding process for the efforts.

### DIFF
--- a/src/util/process_life_cycle.py
+++ b/src/util/process_life_cycle.py
@@ -397,7 +397,6 @@ class ProcessLifeCycle:
         self.fsm.trigger_event(TrdEvent.SHUT_DOWN_ON_ERROR)
         exit_code = ExitCode.GENERAL_ERROR
         if os.path.exists(EXIT_CODE_FILE):
-            logger.info("NOCHEM exit code file exists")
             try:
                 with open(EXIT_CODE_FILE, "r") as f:
                     exit_code_value = int(f.read().strip())
--- a/src/util/process_life_cycle.py
+++ b/src/util/process_life_cycle.py
@@ -3,6 +3,7 @@ import logging
 import queue
 import signal
 import platform
+import os
 from signal import SIGABRT, SIGILL, SIGSEGV, SIGTERM
 
 if platform.system() != "Windows":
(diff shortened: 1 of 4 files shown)


TPXP:
Writing the exit code to a temporary file is a bit odd, can't we use a singleton python module to store this data?

Here's a basic implementation 
```py
# exit_code.py
exitCode = -1
exitOrigin = ""

def set_exit_code(origin: str, code: int):
  if exitOrigin == "":
    exitCode = code
    exitOrigin = origin
  else:
    logger.warn(f"{origin} tried to set exit code to {code} while it had already been by {exitOrigin} to {exitCode}. Ignoring")

def get_exit_code() -> int:
  return exitCode
```

Usage:
```
from exit_code import set_exit_code

# ... function code ...
set_exit_code("module.my_function", 3)
```

src/util/process_life_cycle.py
```diff
@@ -394,6 +395,27 @@ def producer_exit_handler(self, signum, frame):
 
     def shut_down_on_error(self):
         self.fsm.trigger_event(TrdEvent.SHUT_DOWN_ON_ERROR)
+        exit_code = ExitCode.GENERAL_ERROR
+        if os.path.exists(EXIT_CODE_FILE):
+            logger.info("NOCHEM exit code file exists")
```
puppetninja, line 400: remove NOCHEM ?

